PR #104, opened 2014-03-01
add pane switch commands

dougblack:
Implements the following pane switch commands.

`<c-w> h` - focus pane left
`<c-w> l` - focus pane right
`<c-w> j` - focus pane below
`<c-w> k` - focus pane above
`<c-w> <c-p>` - focus previous pane
`<c-w> <c-v>`/`<c-w> v` - create vertical split
`<c-w> <c-s>`/`<c-w> s` - create horizontal split

I manually implemented the focus switches so that the cursor would always be on screen after the switch.

### DIFF
--- a/spec/panes-spec.coffee
+++ b/spec/panes-spec.coffee
@@ -30,7 +30,7 @@ describe "Panes", ->
         getActiveView: -> editorView
       }
 
-    describe "focus pane right", ->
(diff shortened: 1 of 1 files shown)


fogine:
It would be really nice if it would be possible to focus `Tree view` if visible, from adjacent pane.

PR #368, opened 2014-08-01
Remain in visual mode when indenting

jspahrsummers:
This allows indentation keybindings to be repeated without the selection being removed.


mmchaney:
vim does *not* stay in visual mode after indent by default. Is there a way to disable the behavior introduced by this diff?

nathansobo:
You could open a PR with a configuration option for it.

PR #386, opened 2014-08-20
:bug: Fix broken test cases

sknoslo:
These test cases were not running due to indentation errors.
The tests for commands 'dj' and 'dk' now correctly fail.


bronson:
Looks like this issue has been completely fixed.

It's easier to see the changes with whitespace ignored: https://github.com/atom/vim-mode/pull/386/files?w=1

maxbrunsfeld:
@sknoslo Thanks so much for noticing this. These tests were indeed totally broken. I've adapted most of the tests so that they pass now. We can always adjust the assertions if we want to change vim-mode's behavior to match vim's more exactly. One of the tests I had to mark pending because it checks for a vim behavior that we currently don't match at all (`dk` on the first line of the file).

PR #397, opened 2014-09-01
jump motions (ctrl-o / ctrl-i)

garcg:
I've been playing around with implementing [jump motions](http://vimhelp.appspot.com/motion.txt.html#jump-motions), to jump back and forth between previous cursor positions.

It is not finished. Before going any further I would prefer to get some feedback whether this makes sense.

Basically, `atom.workspace.vimState.jumpList` will keep track of the history of cursor positions across buffers in the current workspace. Entries on this list are added when a jump-motion is executed (gg, G, /, and some other commands).
Then ctrl-o, ctrl-i is used to jump to older and newer positions, that could be in the same buffer, or in a different buffer - using atom.open in this case, to switch to the appropriate tab, or open the file if it was closed.

There are a few things pending, like creating an entry in the jumplist when switching buffers/tabs, which I'm not sure how to implement at the moment.

Any feedback on this? ideas?

### DIFF
--- a/docs/motions.md
+++ b/docs/motions.md
@@ -33,3 +33,5 @@
 * [T](http://vimhelp.appspot.com/motion.txt.html#T)
 * [;](http://vimhelp.appspot.com/motion.txt.html#%3B)
 * [,](http://vimhelp.appspot.com/motion.txt.html#%2C)
+* [ctrl-o](http://vimhelp.appspot.com/motion.txt.html#CTRL-O)
+* [ctrl-i](http://vimhelp.appspot.com/motion.txt.html#CTRL-I)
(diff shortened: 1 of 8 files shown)


bronson:
It's too bad this issue never got any comments.

Jump mode (super necessary!) works great on my machine by installing the cursor-history package and sticking this in my keymap:

```cson
'atom-text-editor.vim-mode.command-mode':
  'ctrl-i': 'cursor-history:next'
  'ctrl-o': 'cursor-history:prev'
```

Near as I can tell, it's working 100%.  It would be nice to integrate it into vim-mode but it probably needs a bit more cooking first.

dubiousdavid:
This is what I had to put in my keymap file for this to work:

```
'atom-text-editor.vim-mode.normal-mode':
  'ctrl-i': 'cursor-history:next'
  'ctrl-o': 'cursor-history:prev'
```

Difference being `normal-mode` instead of `command-mode`.

rgkirch:
Thank you for this solution. I'm glad that I have this feature now. I hope that it becomes a default feature soon.

lee-dohm:

PR #416, opened 2014-10-09
Case insensitive search

isaachess:
- If search string all lowercase, defaults to case-insensitive.
- If search string has a single uppercase, defaults to case-sensitive.


morenoh149:
is the same as searching when pressing `/` ? when I search using the forward slash the search is case sensitive but I would like case insensitive.

PR #417, opened 2014-10-13
Fixed yank cursor bug

isaachess:
- When yanking, there was a bug that made the cursor move to the very end of the yank selection, rather than moving the cursor to the beginning of the selection like vim does. After the yank, cursor now moves to start of selection.

- Also, @editor.getSelection() is deprecated, so we are now using editor.getLastSelection().

### DIFF
--- a/lib/operators/general-operators.coffee
+++ b/lib/operators/general-operators.coffee
@@ -141,11 +141,11 @@ class Yank extends Operator
   #
   # Returns nothing.
   execute: (count=1) ->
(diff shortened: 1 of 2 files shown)


bronson:
Looks like the problems addressed by this PR are completely solved in 0.54.0.  At least, viwy and yiw jump to the beginning of the selection (even when selecting in reverse), and `yy` `yj` etc leave the cursor in place.

maxbrunsfeld:
Thanks @bronson. Closing this out.

PR #422, opened 2014-10-16
Don't create VimState if one already exists

erisdev:
Fixes vidius/atom-pain-split#5, among others. Did you know that `eachEditorView` revisits a view when you move it to another pane? Because I had no idea. Very subtle.

### DIFF
--- a/lib/vim-mode.coffee
+++ b/lib/vim-mode.coffee
@@ -17,7 +17,7 @@ module.exports =
       return if editorView.mini
 
       editorView.addClass('vim-mode')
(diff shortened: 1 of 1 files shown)


bronson:
I think this is no longer an issue?

There is a problem where the new pane appears scrolled to the top so the cursor isn't initially onscreen, but a quick scroll shows that it's there.  This isn't specific to vim-mode (I don't think!).

erisdev:
I dunno, I haven't checked on or cared about this since last year. the patch is out of date and my copy of the repository is gone anyway.

PR #432, opened 2014-11-01
optimize find motion visual effect

octoyuz:
I optimize the visual effect of search motion using Atom's decoration flash effect.
https://github.com/atom/atom/blob/v0.141.0/src/decoration.coffee#L162

Since https://github.com/atom/atom/blob/v0.141.0/src/highlight-component.coffee#L42 doesn't provide a callback after completion, I use setTimeout() to destroy the marker.

Following is the screenshot
vimsearch

### DIFF
--- a/lib/motions/search-motion.coffee
+++ b/lib/motions/search-motion.coffee
@@ -24,9 +24,16 @@ class SearchBase extends MotionWithInput
     @
 
   execute: (count=1) ->
+    flashDuration = 500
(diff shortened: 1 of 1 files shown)


bronson:
Similar to discussion in #653, this is probably best implemented by adding a hook to vim-mode, then doing the highlighting in a separate plugin.

lee-dohm:

Issue #188, opened 2014-03-19
cs: vim-surround


lloeki:
vim-surround is non standard but is an incredibly useful plugin by Tim Pope.

Example commands:

`cs"'`: change closest double quotes surrounding the cursor to single quotes
`csw{`: wrap word with spaced curly braces
`csW)`: wrap Word with non-spaced parentheses
`<selection>S[`: wrap selection in spaced square brackets
`cst{`: change surrounding html/xml tag with spaced `{`
`csw<foo>`: surround word with `<foo></foo>`
`csw<foo class='bar'>`: surround word with `<foo class='bar'></foo>`
`ysft)`: wrap current cursor position to next `f` character with non-spaced parentheses

The last ones may be a bit convoluted to implement, especially without real text-object support, but the first four are definitely among the most frequently used ones.

tangledhelix:
AFAICT in the past couple of years, vim-surround has matured and is "done" - I use it all the time in fact. Can this issue be closed?

lee-dohm:
As stated in [the README](https://github.com/atom/vim-mode), this package is no longer maintained and is deprecated. We recommend that people use the [vim-mode-plus package](https://atom.io/packages/vim-mode-plus) instead. Because of this, we are [archiving this repository](https://help.github.com/articles/about-archiving-repositories/) and closing all issues and pull requests. Thanks very much for your support and contributions!

PR #461, opened 2014-12-04
vim-mode extensibility 

saidai-no:
Add interface for external vim-mode extension packages.

Extension packages are intended to augment vim-mode operator suite for some specific areas not appropriate to be included into main vim-mode distribution. Examples of such areas are:

* port of existing Vim plugin (e.g. `vim-surround` #188)
* language-specific motions (e.g. https://github.com/guns/vim-sexp)
* late-binding vim-mode functionality for existing packages (e.g. git integration)

The proposed workflow for extensions is:

* in `.activate` subscribe to `VimMode.observeVimStates`. In callback register new commands (operators/motions) which can use operation stack by means of `pushOperations` method of provided `VimState`
* at `vim-mode@did-deactivate` unregister provided commands and probably self-deactivate

This workflow will be changed when (if) Atom package dependence mechanism is introduced.

`vim-mode::getStateForEditor` is a convenience method which can be useful for just-installed packages.

### DIFF
--- a/lib/vim-mode.coffee
+++ b/lib/vim-mode.coffee
@@ -35,12 +35,11 @@ module.exports =
       @editorStateMap[editor.id] = vimState
       @emitter.emit 'did-attach', vimState
 
-      stateDisposable = new Disposable =>
-        delete @editorStateMap[vimState.editor.id]
+      stateDisposable = new Disposable ->
         vimState.destroy()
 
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Most of this has been implemented in #540. Feel free to open new issues for adding to the API that has been introduced.

Issue #168, opened 2014-03-13
Add support for `_` as an alternative to `^`


lloeki:
Useful for layouts where `^` is a dead key.

lee-dohm:
As stated in [the README](https://github.com/atom/vim-mode), this package is no longer maintained and is deprecated. We recommend that people use the [vim-mode-plus package](https://atom.io/packages/vim-mode-plus) instead. Because of this, we are [archiving this repository](https://help.github.com/articles/about-archiving-repositories/) and closing all issues and pull requests. Thanks very much for your support and contributions!

PR #467, opened 2014-12-09
Underscore support to move to front of line

ftwillms:
Related to this enhancement: https://github.com/atom/vim-mode/issues/168

I'm not too familiar with the spec testing and plan to add that, but I wanted to go ahead and send this pull request to see how off the mark this implementation might be. Originally, it was just going to be a extra mapping of the '^' key, but I learned something new from the issue comments which is that '_' with the prefix count will move it down as well. This is also my first public pull requests, so forgive me for my ignorance.


maxbrunsfeld:
:+1: I'll definitely merge if you add some specs.

ftwillms:
Added the spec and while I was at it I noticed other, similar motions have since been updated (I'm guessing due to the Atom API release) so I've mimicked this new style as well.

maxbrunsfeld:
Great work @ftwillms thanks a lot.

Issue #388, opened 2014-08-21
ESC in non-command mode kills multiple cursors


brandonhorst:
When you have multiple cursors open, tapping escape to bring yourself out back to command mode shouldn't also kill your cursors, so you can continue working with them.

However, ESC in command mode should kill your cursors.

PR #469, opened 2014-12-10
Escape only clears multiple cursors in command mode

brandonhorst:
Fixes https://github.com/atom/vim-mode/issues/388


maxbrunsfeld:
This ended up getting fixed via #477. Sorry for the slow response; see my comment on [your other PR](https://github.com/atom/vim-mode/pull/473).

PR #471, opened 2014-12-13
[RFC] Defer vim-state.coffee loading

saidai-no:
I was wondering is it acceptable to defer vim-state.coffee loading until the first text editor is created?

On my machine with this change vim-mode loading time goes from ~100ms down to ~15ms.
On the other hand, this change will slow down first text editor opening a bit.

Change was proposed by @lee-dohm in atom/atom#2654.

### DIFF
--- a/lib/vim-mode.coffee
+++ b/lib/vim-mode.coffee
@@ -1,7 +1,7 @@
 {Disposable, CompositeDisposable} = require 'event-kit'
 StatusBarManager = require './status-bar-manager'
 GlobalVimState = require './global-vim-state'
-VimState = require './vim-state'
+VimState = null
 
 module.exports =
   config:
@@ -20,6 +20,7 @@ module.exports =
     @disposables.add statusBarManager.initialize()
     @disposables.add atom.workspace.observeTextEditors (editor) =>
       return if editor.mini
+      VimState ?= require './vim-state'
 
       element = atom.views.getView(editor)
 


maxbrunsfeld:
In my opinion, since this won't help the common case of starting Atom with text editors open, it isn't worth lazy-loading in this case. I'm open to arguments in favor of doing it though.

bronson:
Things have gotten a fair bit faster since this PR was filed.  I figured I'd run this test again...

#### Test

* `git clone bronson/vim-mode` # (to get the rebased [lazy loading](https://github.com/bronson/vim-mode/commits/lazy-loading) branch)
* `apm link vim-mode`

Now, to perform the test, I launched Atom, opened timecop, and recorded vim-mode's penalty.  I ran the following commands alternately so each startup was very warm.

* `git checkout lazy-loading; sleep 0.5; atom .`
* `git checkout master; sleep 0.5; atom .`

#### Results

|Lazy|Master|
|------|-------|
|19ms|30ms|
|18ms|30ms|
|18ms|31ms|

Stable and definitive: on my machine, this technique saves 10ms / ~30%.

#### Conclusion

18ms is worthy of a gold `highlight-warning` and 30ms is the red `highlight-error`.  This PR might get vim-mode out of the status-bar and symbols-tree-view doghouse.

That said, I agree with @maxbrunsfeld: every time I open my editor, I'm going straight into a Vim window.  For me, all this patch does is reorder the load process.  It doesn't actually save any real-world time.  Doesn't seem worth it.

maxbrunsfeld:
Thanks @bronson. I think we'll need to use some different tactics for improving the package's load time. Closing this out.

PR #472, opened 2014-12-13
Fix scrolling up/down so it moves by 1 line.

### LINKED ISSUE #430: Scroll window down command moves by 3 lines not 1
The [scroll window down command](http://vimhelp.appspot.com/scroll.txt.html#CTRL-E) moves the screen by 3 lines not 1. [Scroll window up](http://vimhelp.appspot.com/scroll.txt.html#CTRL-Y) works as expected.

neojski:
Fix #430 once https://github.com/atom/atom/pull/4597 is merged.

Unfortunately, atom is not showing the integer number of lines all the time
so sometimes it looks like it's not scrolling correctly. This is because
scrolling up (ctrl-y) is done by requesting atom to show the top lines and
so very bottom line may be only partially shown (and so it sometimes feels
like it does not work correctly).

### DIFF
--- a/lib/scroll.coffee
+++ b/lib/scroll.coffee
@@ -5,7 +5,7 @@ class Scroll
     @scrolloff = 2 # atom default
     @rows =
       first: @editor.getFirstVisibleScreenRow()
-      last: @editor.getLastVisibleScreenRow()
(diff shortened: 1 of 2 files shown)


neojski:
See #5951

Issue #465, opened 2014-12-08
Cannot exit visual mode by hitting 'v' key


learning:
Cannot exit visual mode by hitting `v` key

liyaodong:
I meet same problem , must hitting `escape` to exit visual mode .

t9md:
Sorry for this mess, I didn't noticed this mess, just wanted to fix my commit log, sorry I should learn git.

PR #476, opened 2014-12-28
:bug: #465: cannot exit visual mode by hitting 'v'

fotanus:
This patch simply maps the 'v' and 'V' keys to reset command mode.
The cursor is kept on the end of the selection, as expected in vim.

### DIFF
--- a/keymaps/vim-mode.cson
+++ b/keymaps/vim-mode.cson
@@ -222,3 +222,5 @@
 '.editor.vim-mode.visual-mode':
   'x': 'vim-mode:delete'
   's': 'vim-mode:change'
+  'v': 'vim-mode:reset-command-mode'
+  'V': 'vim-mode:reset-command-mode'


maxbrunsfeld:
I'd like to merge this, but with just `v`, not the `V` mapping. In regular visual mode, `V` should actually switch to *linewise* visual mode. That behavior isn't implemented yet, but in the meantime I'd prefer not to add a mapping for `V`.

fotanus:
I'm actually using `V` [since it is already mapped](https://github.com/atom/vim-mode/blob/master/keymaps/vim-mode.cson#L141). Of course it lacks many crucial features, but since it is there already at least to select code, I don't see why not merging this. It is helpful to select code line by line, and I wouldn't be surprise people would notice if we simply remove the current V mode support. What do you think?

bronson:
This functionality appears to be in vim-mode now.  `v` enters visual mode, `v` again exits and leaves the cursor at the last position.

In addition, @maxbrunsfeld's request appears to work too.  `v` starts plain visual, `V` converts to linewise, and `v` back to regular.  Cursor remains in the correct position at all times.

There's one minor bug that I can tell, probably worth opening a new issue for: put the cursor in the middle of the line, hit V to start linewise, then V to end.  The cursor ends up at the end of the line, but in Vim it doesn't move.

But I think this PR is completely addressed?

t9md:
#695 include fix , and at merged related issue was closed. Maybe @maxbrunsfeld forgot to close?

maxbrunsfeld:
> Maybe @maxbrunsfeld forgot to close this one?

:+1:

PR #506, opened 2015-02-05
showing prefix char in command mode input view

coolwanglu:
This patch adds a 'prefix' to the command mode input when necessary. 
And currently only the `/` motion is supported, any other motion we should support?

When '/' is pressed, '/' is shown in the beginning of the input, and when '/' is removed, the motion is cancelled naturally.

But this is not perfect, since we can select & paste the prefix, which is not possible in vim. We might want to disable selection of the command mode input.

I think with this, we can easily support some command-line mode commands (`:` commands)?

### DIFF
--- a/lib/motions/search-motion.coffee
+++ b/lib/motions/search-motion.coffee
@@ -8,10 +8,10 @@ class SearchBase extends MotionWithInput
   operatesInclusively: false
   @currentSearch: null
 
-  constructor: (@editor, @vimState) ->
+  constructor: (@editor, @vimState, @reverse=false) ->
     super(@editor, @vimState)
     Search.currentSearch = @
-    @reverse = @initiallyReversed = false
+    @initiallyReversed = @reverse
 
   repeat: (opts = {}) =>
     reverse = opts.backwards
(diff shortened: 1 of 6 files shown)


coolwanglu:
Oh this conflicts with search history.
I'm not pretty sure about how histories are organized in native vim, but we may cache all the commands and search the history through prefixes. Which is also good for the future support for `:` commands.

coolwanglu:
Also we might want to prevent the cursor being moved to the beginning when prefixChar is presented.

coolwanglu:
Now all specs are passed.
Please review.

bronson:
I resurrected this branch to evaluate it: https://github.com/bronson/vim-mode/commits/prefix-char

It looks like https://github.com/atom/vim-mode/pull/582 has blown it away.  It would probably be easier to rewrite the patch than to resolve the merge conflicts.

I don't understand why this was never reviewed.  :(

ardrigh:
It would be nice if these features could be reviewed along with what https://github.com/jazzpi/ex-mode is doing with the commands. I added a similar request to that package to add <kbd>:</kbd> to the start of the command input.

If vim-mode and ex-mode merged together, then these requests would be better handled IMO.

lee-dohm:

PR #510, opened 2015-02-06
beep when search history is exhausted

coolwanglu:


coolwanglu:
Do we want beeping in vim-mode? I could add lots more (moving up over the first line etc).

jeffschwartz:
I'd only like it if it is an option and I can turn it off via options. Otherwise I'd view it as intrusive. No offense intended of course :)

coolwanglu:
@jeejkang  There's an option for atom to turn of beeping globally. But of course we can make a vim-mode-only option.

bronson:
I don't quite get it...  Stock vim doesn't beep when you've exhausted search history (renamed my ~/.vim directory and launched with `vim -u /dev/null` so I think it's stock).

Atom doesn't seem to either.  (hitting cmd-f, then up-arrow until it stops...  no beep)

The patch looks great but is there a precedent for this feature?

lee-dohm:

PR #524, opened 2015-02-15
:bug: fix tab handling in shift-B

jacekkopecky:
shift-B would currently only look for spaces and newlines; it needs to handle any whitespace in skipping to the previous word

AllWhitespace modeled after lib/text-objects.coffee

This is my first little contribution to Atom, feel free to let me know if I'm doing something wrong.


maxbrunsfeld:
Awesome, thanks!

Issue #308, opened 2014-06-07
Copy to clipboard with "y"


fotanus:
Hi,

This is not the current default, but I think it makes sense. It can [easily be done](https://atom.io/docs/api/v0.101.0/api/classes/Editor.html#copySelectedText-instance). I would be happy to provide a pull request.

In case this is not a good idea, what is the current way we have to copy things to clipboard? I failed in find it.

PR #525, opened 2015-02-15
Adds useClipboardAsDefaultRegister option.

chrisfarms:
I find it much nicer to have the default register mapped to the clipboard for better interop with the standard editor world, so this PR adds a new option to enable that behaviour.

It also moves the settings into their own module to serve a single point for defining options and any associated lookup functions.

Related:  #308

### DIFF
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -1,5 +1,6 @@
 _ = require 'underscore-plus'
 {Point, Range} = require 'atom'
+settings = require '../settings'
 
 WholeWordRegex = /\S+/
 WholeWordOrEmptyLineRegex = /^\s*$|\S+/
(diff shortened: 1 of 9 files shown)


maxbrunsfeld:
I absolutely agree; I've meant to add this for a while. Would you mind adding tests for this case? I think it would be sufficient to add one test to [the describe for 'y'](https://github.com/atom/vim-mode/blob/master/spec/operators-spec.coffee#L533) and one test to [the describe for 'p'](https://github.com/atom/vim-mode/blob/master/spec/operators-spec.coffee#L730), each of which says something like:

```coffee
describe "when the 'useClipboardAsDefaultRegister' settings is set to true", ->
  it "copies to the clipboard", ->
   spyOn(atom.clipboard, "write")
   keydown("y")
   expect(atom.clipboard.write).toHaveBeenCalledWith("something...")
```

chrisfarms:
np.... added tests and rebased against master.

maxbrunsfeld:
Perfect. Thanks a lot!

hanchang:
Is there a way to disable this via config? I guess I'm one of the weird ones that doesn't like my OS system clipboard to be the same as my default `vim` clipboard register. Thanks in advance!

maxbrunsfeld:
Yeah, it should appear in the settings view under the vim mode package.

hanchang:
Amazing, thanks!

PR #526, opened 2015-02-15
feature: VIM whichwrap

jacekkopecky:
Hi, 
VIM has the `whichwrap` setting that allows wrapping some operations across line boundaries. I have now added wrapping of left and right motion (VIM values h,l,<,>) for myself because I use it heavily; I'm submitting it here for your consideration.

I know that this particular setting for `h` and `l` is "not recommended" in VIM, but the reasons for it that I could find are to do with plugins expecting the non-wrapping behavior – does anybody here know of actual problems with wrapping `h`,`l`, especially if they would translate to vim-mode in Atom?

Note that the VIM whichwrap settings `b`, `[` and `]` are already the standard behaviour in Atom vim-mode. The remaining settings `s` and `~` I don't really care about.

### DIFF
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -150,7 +150,7 @@ class MoveRight extends Motion
   moveCursor: (cursor, count=1) ->
     _.times count, =>
       cursor.moveRight() unless cursor.isAtEndOfLine()
-      cursor.moveRight() if cursor.isAtEndOfLine() and atom.config.get('vim-mode.wrapLeftRightMotion')
+      cursor.moveRight() if atom.config.get('vim-mode.wrapLeftRightMotion') and cursor.isAtEndOfLine()
       @ensureCursorIsWithinLine(cursor)
 
 class MoveUp extends Motion
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -141,7 +141,7 @@ class MoveLeft extends Motion
 
   moveCursor: (cursor, count=1) ->
     _.times count, =>
-      cursor.moveLeft() unless cursor.isAtBeginningOfLine() and !atom.config.get('vim-mode.wrapLeftRightMotion')
+      cursor.moveLeft() unless cursor.isAtBeginningOfLine() and not atom.config.get('vim-mode.wrapLeftRightMotion')
       @ensureCursorIsWithinLine(cursor)
 
 class MoveRight extends Motion
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -141,7 +141,7 @@ class MoveLeft extends Motion
 
   moveCursor: (cursor, count=1) ->
     _.times count, =>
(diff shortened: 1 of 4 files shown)


lib/motions/general-motions.coffee
```diff
@@ -150,6 +150,7 @@ class MoveRight extends Motion
   moveCursor: (cursor, count=1) ->
     _.times count, =>
       cursor.moveRight() unless cursor.isAtEndOfLine()
+      cursor.moveRight() if atom.config.get('vim-mode.wrapLeftRightMotion') and cursor.isAtEndOfLine()
```
maxbrunsfeld: Does `l` still work with this change? It used to call `cursor.moveRight()` *unless* the cursor was at the end of the line. Now, if `wrapLeftRightMotion` is `true`, it does it *only if* the cursor is at the end of the line.

I would think it should be this:

```coffee
cursor.moveRight() if (not cursor.isAtEndOfLine()) or atom.config.get('vim-mode.wrapLeftRightMotion')
```

lib/motions/general-motions.coffee
```diff
@@ -140,9 +140,9 @@ class MoveLeft extends Motion
   operatesInclusively: false
 
   moveCursor: (cursor, count=1) ->
-    _.times count, ->
-      unless cursor.isAtBeginningOfLine()
-        cursor.moveLeft()
+    _.times count, =>
+      cursor.moveLeft() unless cursor.isAtBeginningOfLine() and !atom.config.get('vim-mode.wrapLeftRightMotion')
```
maxbrunsfeld: This looks right, but the combination of `unless` and `!` is a little confusing to me. Could you switch this to use an `if` instead?

lib/motions/general-motions.coffee
jacekkopecky: `l` is actually called once on the line before, and then, if we're at the end of the line, it's called again so Atom takes care of going to the next line for us.

maxbrunsfeld: Got it :+1:

lib/motions/general-motions.coffee
jacekkopecky: Sorry, I'm not sure I can do that. I could make it 
```
      cursor.moveLeft() if atom.config.get('vim-mode.wrapLeftRightMotion') or !cursor.isAtBeginningOfLine()
or
      cursor.moveLeft() if !cursor.isAtBeginningOfLine() or atom.config.get('vim-mode.wrapLeftRightMotion')
```
But that's not really better, I suspect. What do you think?

maxbrunsfeld: Either is fine with me, except that we generally use coffee-script's `not` instead of `!`.

jacekkopecky: thanks, I'm new to coffeescript. Sorry about the double commit, I missed part of your reply.

maxbrunsfeld: No worries! And the `not` thing is just our preference in the Atom codebases; not something I'd expect anyone to just *know* as contributor. Thanks a lot for all of these contributions.

PR #527, opened 2015-02-17
Fix repeated find motion

jacekkopecky:
There was a bug that repeated f/t motion wouldn't get canceled if we tried to move too many times.

E.g. in the test string 'abcabcabcabc' with cursor at the start, the motion 11fa would actually jump to the second a; the old test didn't uncover it because, as it happened, 10fa jumped to the first 'a', so the cursor didn't move. It was because indexOf with a negative fromIndex simply searched the whole string again.

Then while testing the backward motion F/T, I found that if it so happens that the character we're looking for is at the beginning of the line, the motion would simply go there, and again wouldn't be canceled. That's because in lastIndexOf, a negative fromIndex is treated as 0 and the first character gets searched (where it shouldn't).

So here I fix both cases. Feel free to split the tests that have accumulated there.

### DIFF
--- a/lib/motions/find-motion.coffee
+++ b/lib/motions/find-motion.coffee
@@ -17,6 +17,7 @@ class Find extends MotionWithInput
     if @backwards
       index = currentPosition.column
       for i in [0..count-1]
+        return if index <= 0 # we can't move backwards any further, quick return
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Thanks!

PR #534, opened 2015-02-21
Adds keymap for insert mode CTRL-U

nicolaiskogheim:
Maps i_CTRL-U to `editor:delete-to-beginning-of-line`

I will write a test for this tomorrow. :smile:


nicolaiskogheim:
No, this should probably not have a test.
Like the `i_CTRL-W` binding (defined the line above my addition), this one just maps to a command defined on `editor`.
That is probably why @maxbrunsfeld didn't add a test for that either.

maxbrunsfeld:
Thanks!

PR #539, opened 2015-02-23
Adds keymap for insert mode CTRL-T and CTRL-D

nicolaiskogheim:
These are standard vim mappings.

`ctrl-d` shadows `core:delete` (delete character to the right of cursor)
`ctrl-t` shadows `editor:transpose` (swaps character to the left of cursor with character on the right)

There are other ways to do the things that get shadowed that is as quick and easy in Vim-mode, but I understand if your reasoning behind this package conflicts with these two additions.

Again, no tests since these are only mappings and not new functionality.

### DIFF
--- a/keymaps/vim-mode.cson
+++ b/keymaps/vim-mode.cson
@@ -13,6 +13,7 @@
 'atom-text-editor.vim-mode.insert-mode':
   'ctrl-w': 'editor:delete-to-beginning-of-word'
   'ctrl-u': 'editor:delete-to-beginning-of-line'
+  'ctrl-t': 'editor:indent-selected-rows'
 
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
@nicolaiskogheim I'm all for adding every last visual and normal mode key-mapping that we can, but this package's approach is to make insert mode behave as much as possible like plain Atom, at least by default.

I'm ok w/ adding settings to enable insert mode features that conflict w/ Atom's, but since these are just mappings, I'd prefer to let people add them in their own `keymap.cson`.

nicolaiskogheim:
All good :+1: 

A setting would've been very nice, so that it would've been possible to chose between sort of atom-with-vim-mode and vim-with-atom-around-it-mode.

Issue #536, opened 2015-02-23
Expose VimState for extensibility


lloeki:
Having read access to an editor's `VimState` is bound to be very useful for third-partyvim-mode extensions such as [ex-mode](https://github.com/lloeki/ex-mode). Currently, this is completely prevented, even in the most hackish ways, because the references are unreachable, as they're trapped into a closure scope:

```
--- lib/vim-mode.coffee.orig	2015-02-23 11:04:48.000000000 +0100
+++ lib/vim-mode.coffee	2015-02-23 11:04:58.000000000 +0100
@@ -9,9 +9,9 @@
 
   activate: (state) ->
     @disposables = new CompositeDisposable
-    globalVimState = new GlobalVimState
+    @globalVimState = new GlobalVimState
     @statusBarManager = new StatusBarManager
-    vimStates = new WeakMap
+    @vimStates = new WeakMap
 
     @disposables.add atom.workspace.observeTextEditors (editor) =>
       return if editor.mini
```

Subsequently, `ex-mode` will be able to query an editor's VimState on its own:

```
VimMode = atom.packages.activePackages['vim-mode'].mainModule
VimMode.globalVimState
VimMode.vimStates.get(atom.workspace.getActiveEditor())
```

Alternatives such as implementing accessor methods (such as `VimMode.getGlobalState()` and `VimMode.getVimState(editor)`, possibly cloning the state objects to guarantee read-only-ness) or going full steam ahead and providing a versioned service, like [atom/status-bar](https://github.com/atom/status-bar/blob/master/package.json#L16) are equally welcome.

maxbrunsfeld:
Yeah, we need to provide a versioned service. I'd be open to a PR that adds one if you're interested in working on that.

lloeki:
I'm on it.

maxbrunsfeld:
Closed via #540.

PR #540, opened 2015-02-24
Publish vim-mode as an atom service

lloeki:
In response to #536.

This exposes `getState(editor)` along with `getGlobalState()` via a versioned `vim-mode` service. The consumer then has direct access to the editor's VimState instance. Therefore I made the service version equal to vim-mode's package version.

This will allow packages extending vim-mode to have access to its core state (such as registers, marks...) but also `pushOperations`, which offers a wealth of possibilities.

It may be advisable ultimately to box the VimState instance into a service proxy object, so as to expose a restricted and stable API, isolated from what amounts to the core of vim-mode. This may be too early though, as we may be too restrictive early on. I don't know what your versioning policy is, but this may be a worthy goal for vim-mode 1.0.

### DIFF
--- a/lib/vim-mode.coffee
+++ b/lib/vim-mode.coffee
@@ -9,23 +9,23 @@ module.exports =
 
   activate: (state) ->
     @disposables = new CompositeDisposable
-    globalVimState = new GlobalVimState
+    @globalVimState = new GlobalVimState
     @statusBarManager = new StatusBarManager
-    vimStates = new WeakMap
+    @vimStates = new WeakMap
 
     @disposables.add atom.workspace.observeTextEditors (editor) =>
       return if editor.mini
 
       element = atom.views.getView(editor)
 
-      if not vimStates.get(editor)
+      if not @vimStates.get(editor)
         vimState = new VimState(
           element,
           @statusBarManager,
-          globalVimState
+          @globalVimState
         )
 
-        vimStates.set(editor, vimState)
+        @vimStates.set(editor, vimState)
 
         @disposables.add new Disposable =>
           vimState.destroy()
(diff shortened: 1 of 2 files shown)


lib/vim-mode.coffee
```diff
@@ -33,8 +33,18 @@ module.exports =
...
+
   consumeStatusBar: (statusBar) ->
     @statusBarManager.initialize(statusBar)
     @statusBarManager.attach()
     @disposables.add new Disposable =>
       @statusBarManager.detach()
+
+  provideVimMode: ->
+    getGlobalState: @getGlobalState.bind(@)
+    getState: @getState.bind(@)
```
maxbrunsfeld: I think I'd rather have this be called `getEditorState` to make it clear that you need to pass a `TextEditor`. You cool w/ that @lloeki?

package.json
```diff
@@ -19,5 +19,13 @@
         "^0.58.0": "consumeStatusBar"
       }
     }
+  },
+  "providedServices": {
+    "vim-mode": {
+      "description": "",
+      "versions": {
+        "0.33.0": "provideVimMode"
```
maxbrunsfeld: In general, we don't intent to have these service versions match the package versions, since the package may go through many versions w/o changing its API. I believe we made this mistake in the status-bar service, which made for a confusing example.

Would you mind changing this to `0.1.0`? I think that is a good default to use, since the service is still new and should be considered somewhat unstable/evolving.

maxbrunsfeld:
Left some minor comments. Otherwise, looks good :+1:.

maxbrunsfeld:
> This may be too early though, as we may be too restrictive early on. I don't know what your versioning policy is, but this may be a worthy goal for vim-mode 1.0.

Sounds reasonable to me.

lib/vim-mode.coffee
lloeki: No problem, I wasn't sure about the name either.

package.json
lloeki: That's sensible.

PR #545, opened 2015-02-28
Use space-pen dependencies for input views

### LINKED ISSUE #544: Search text not displaying using /
Since the latest update I'm not seeing search text using the vim-mode search (`/`). The search functions correctly, but the search box doesn't show the text I'm typing:

image

Atom 0.184, vim-mode 0.34, osx

MattKunze:
This probably isn't the best way to fix this, the views seem to be deprecated. Not sure how to go about replacing the views with an `<atom-text-editor>` though

fixes #544

### DIFF
--- a/lib/view-models/vim-command-mode-input-view.coffee
+++ b/lib/view-models/vim-command-mode-input-view.coffee
@@ -1,4 +1,4 @@
-{View} = require 'space-pen'
+{View, $} = require 'space-pen'
 {TextEditorView} = require 'atom-space-pen-views'
 
 module.exports =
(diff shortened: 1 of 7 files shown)


MattKunze:
Doh, seems to fail a bunch of tests somehow. Not sure if it makes more sense to update the tests against the deprecated views, or update how the input editor view works

MattKunze:
Most of the failing tests seem to stem from needing to change how events are bound to the text editor now that it uses shadow DOM. I can update the bindings with

```coffee
  handleEvents: ->
    input = $ @editor[0].rootElement.querySelector 'input'
    if @singleChar?
      input.on 'textInput', @autosubmit
    @editor.on 'core:confirm', @confirm
    @editor.on 'core:cancel', @cancel
    input.on 'blur', @cancel
```

and single character motions are working again, but the specs still are failing

MattKunze:
@maxbrunsfeld - I was able to return to this and get the tests to work, the specs had the same issue getting a reference to the text editor input through the shadow DOM mentioned above.

It seems like there has to be a better way to hook to the input, but I couldn't find it in the docs. I'd be happy to update if someone would point me in the right direction

tmm1:
Issue is fixed on master.

maxbrunsfeld:
Hey @MattKunze, didn't mean to ignore this. @tmm1 and I ended up fixing this a different way with the added benefit of not relying on jQuery any more.

MattKunze:
No worries, just glad to see the issue fixed!

Issue #418, opened 2014-10-13
Change inner tag not working


isaachess:
`cit` and `dit` should work with vim tag-blocks, but are not implemented.

bradleyflood:
+1 for `cit` and `dit` which I use very frequently as a front-end dev.

jeffschwartz:
+ 1

joshmedeski:
+1!

jeffschwartz:
+1

neiled:
:+1:

maxbrunsfeld:
This has been implemented in #547. It doesn't exactly match vim's behavior but it's pretty good.

bradleyflood:
Excellent thank you

isaachess:
@maxbrunsfeld No offense, but I greatly disagree -- currently implementation of change-inner-tag is basically useless, doesn't work like vim at all. I would rather have it not implemented than its current state, where the behavior feels pretty unpredictable.

Either way, I think we should still have an open issue for this, to get it to match vim's actual feature.

maxbrunsfeld:
@isaachess sounds great. Feel free to open an issue.

collegeimprovements:
:+1:  for `cit` and `dit`. It would also be great if for `ruby` files, you can implement method navigation `]m` and selection `vim` bindings.

PR #547, opened 2015-03-03
Add 'i t' for inside tags

neiled:
I just :+1: #418 and thought I'd give a go at actually implementing it.

Hopefully code is straight forward.
It's not exactly like vim right now, if you're not inside a tag it won't do anything, vim is smart enough to move to the next tag and then do the action. However that seems to be how the other motions in atom are done right now anyway...

### DIFF
--- a/keymaps/vim-mode.cson
+++ b/keymaps/vim-mode.cson
@@ -201,6 +201,7 @@
   'i B': 'vim-mode:select-inside-curly-brackets'
   'i <': 'vim-mode:select-inside-angle-brackets'
   'i >': 'vim-mode:select-inside-angle-brackets'
+  'i t': 'vim-mode:select-inside-tags'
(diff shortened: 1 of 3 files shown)


murphyrandle:
@neiled Funny, I was just about to take another crack at this myself. I started a while back, but got all hung up on handling cases where `<` and `>` are used inside tags, but aren't tags themselves. But honestly, I'd much rather have the feature since it works in most cases than worry about that edge case.

murphyrandle:
Also, super great profile picture :+1:

maxbrunsfeld:
Yeah, nice idea @neiled. It seems like this covers the majority of the use cases for `i t`. We can always make it fancier later.

neiled:
Great! This might close #418 too perhaps @maxbrunsfeld?

neiled:
And thanks @murphyrandle :wink:

PR #548, opened 2015-03-04
:bug: fix 'd' and 'c' sometimes deleting a char before cursor

### LINKED ISSUE #543: 'ct' deletes one character before the cursor
I thought there is already an issue for that, but I can't find it anymore. Sorry, if this is a duplicate.

In some situations the commands `cf`, `ct`, `dt` and `df` also delete one character before the cursor. I don't know the exact circumstances , but could reproduce it as follow:

1. Insert a new line `test (xyz)` and leave insert mode.
2. Yank and put the line after the inserted line: `yyp`
3. Go to x: `fx`
4. Change til closing bracket: `ct)`
5. Result: `test )`

I'm using Atom 0.182.0 and vim-mode 0.34.0 on Linux (x64).

deiwin:
This fixes #543

For some reason having yanked the whole line before the delete + motion made the wasReversed variable be true (it was false without the yank)

### DIFF
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -82,7 +82,7 @@ class Motion
 
       if (isReversed or isEmpty) and not (wasReversed or wasEmpty)
         selection.setBufferRange([newStart, [newEnd.row, oldStart.column + 1]])
-      if wasReversed and not isReversed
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
:clap: Wow, that was some really nice debugging @deiwin. Thank you very much!

PR #553, opened 2015-03-08
Update to atom 1.0 api (fixes #552)

### LINKED ISSUE #552: Cannot see text input during search

lightsofapollo:
I am usually a heavy (m)vim user but today I tried to play with atom and generally like it... The big bug I hit was not being able to see my search input as I typed.. I am only somewhat confident this makes sense but it looks right given atom core code/deprecation warnings.

### DIFF
--- a/lib/view-models/vim-command-mode-input-view.coffee
+++ b/lib/view-models/vim-command-mode-input-view.coffee
@@ -1,4 +1,4 @@
-{View, TextEditorView} = require 'atom'
+{View, TextEditorView} = require 'atom-space-pen-views'
 
 module.exports =
 class VimCommandModeInputView extends View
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Thanks for the PR! It looks like some of the tests are failing. I'll try to look into why when I get a chance.

MattKunze:
#545 has the same changes and I just got the tests to work FWIW. It seems a little ugly how it's traversing the text editor component though

bronson:
Looks like this was fixed by fc24df5be327b1ff7643f648c5f68910231cc7a4?

maxbrunsfeld:
Yeah, this has been taken care of now. Thanks for looking into it though, @lightsofapollo.

PR #555, opened 2015-03-08
fix wrapping behavior of x

jacekkopecky:
`x` should only wrap to next line with `vim-mode.wrapLeftRightMotion = true`, and only with a count or on an empty line.
The PR also adds a bit to the tests for X.

### DIFF
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -149,10 +149,16 @@ class MoveLeft extends Motion
 class MoveRight extends Motion
   operatesInclusively: false
 
-  moveCursor: (cursor, count=1) ->
+  # MoveRight can be combined with Delete, which will indicate it by the presence of options.deleting
+  moveCursor: (cursor, count=1, options={}) ->
     _.times count, =>
-      cursor.moveRight() unless cursor.isAtEndOfLine()
-      cursor.moveRight() if settings.wrapLeftRightMotion() and cursor.isAtEndOfLine()
+      goingToNextLine = atom.config.get('vim-mode.wrapLeftRightMotion')
+
+      if not cursor.isAtEndOfLine() # the line is not empty
+        cursor.moveRight()
+        goingToNextLine = false if options.deleting # when deleting, wrapLeftRightMotion only works on an empty line
+
+      cursor.moveRight() if cursor.isAtEndOfLine() and goingToNextLine
       @ensureCursorIsWithinLine(cursor)
 
 class MoveUp extends Motion
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -149,16 +149,16 @@ class MoveLeft extends Motion
 class MoveRight extends Motion
   operatesInclusively: false
 
-  # MoveRight can be combined with Delete, which will indicate it by the presence of options.deleting
   moveCursor: (cursor, count=1, options={}) ->
     _.times count, =>
-      goingToNextLine = settings.wrapLeftRightMotion()
+      wrapToNextLine = settings.wrapLeftRightMotion()
 
-      if not cursor.isAtEndOfLine() # the line is not empty
-        cursor.moveRight()
-        goingToNextLine = false if options.deleting # when deleting, wrapLeftRightMotion only works on an empty line
+      # when the motion is combined with an operator, we will only wrap to the next line
+      # if we are already at the end of the line (after the last character)
+      wrapToNextLine = false if @vimState.mode is 'operator-pending' and not cursor.isAtEndOfLine()
 
-      cursor.moveRight() if cursor.isAtEndOfLine() and goingToNextLine
+      cursor.moveRight() unless cursor.isAtEndOfLine()
+      cursor.moveRight() if wrapToNextLine and cursor.isAtEndOfLine()
       @ensureCursorIsWithinLine(cursor)
 
 class MoveUp extends Motion
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -149,7 +149,7 @@ class MoveLeft extends Motion
 class MoveRight extends Motion
   operatesInclusively: false
 
-  moveCursor: (cursor, count=1, options={}) ->
+  moveCursor: (cursor, count=1) ->
     _.times count, =>
       wrapToNextLine = settings.wrapLeftRightMotion()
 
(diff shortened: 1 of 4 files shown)


lib/motions/general-motions.coffee
```diff
@@ -149,10 +149,16 @@ class MoveLeft extends Motion
...
+  # MoveRight can be combined with Delete, which will indicate it by the presence of options.deleting
+  moveCursor: (cursor, count=1, options={}) ->
     _.times count, =>
-      cursor.moveRight() unless cursor.isAtEndOfLine()
-      cursor.moveRight() if settings.wrapLeftRightMotion() and cursor.isAtEndOfLine()
+      goingToNextLine = atom.config.get('vim-mode.wrapLeftRightMotion')
+
+      if not cursor.isAtEndOfLine() # the line is not empty
+        cursor.moveRight()
+        goingToNextLine = false if options.deleting # when deleting, wrapLeftRightMotion only works on an empty line
```
maxbrunsfeld: I'm pretty sure `h` should perform this motion for *any* operator, not just `d` (e.g. `y`, `gU`, `g~`). Could you try checking `@vimState.mode is 'operator-pending'` instead of `options.deleting`? If that works, you wouldn't need the `deleting` option that you added on `general-operators.coffee:87`, and could probably :fire: these comments :point_up:.

spec/operators-spec.coffee
```diff
@@ -39,65 +39,179 @@ describe "Operators", ->
...
+          keydown('x')
+          expect(editor.getText()).toBe 'a0123\n\nxyz'
+          expect(editor.getCursorScreenPosition()).toEqual [0, 1]
+          expect(vimState.getRegister('"').text).toBe 'bc\n'
+
+          keydown('7')
+          keydown('x')
+          expect(editor.getText()).toBe 'ayz'
+          expect(editor.getCursorScreenPosition()).toEqual [0, 1]
+          expect(vimState.getRegister('"').text).toBe '0123\n\nx'
```
maxbrunsfeld: Nice tests :sparkles:.

lib/motions/general-motions.coffee
jacekkopecky: Oh right, I'll check that - it would be cleaner indeed. I just need to ask for patience please, my dayjob is extra busy right now so I can't give vim-mode as much time as I'd like... Thank you.

jacekkopecky: Just checked it and it's not simple:
 - `gUl` will not move the cursor in VIM, but in vim-mode it does (I may look into that)
 - `gUl` will do nothing in VIM on an empty line, in vim-mode with wrapping on it will move to the next line (and probably try to uppercase the newline which is a no-op)
 - `gU3l` in VIM will span a newline if done on an empty line or near the end of a line, so does vim-mode but it uppercases one more character than VIM does, in a way that's more consistent with the `l` motion so it doesn't bother me

 - `yl` in VIM on a non-empty line only yanks the character under the cursor, same in vim-mode
 - `yl` in VIM on an empty line yanks the newline and treats it as linewise for `p`; vim-mode does the same 
 - `y2l` in VIM on a non-empty line on the last character only yanks that character, not including the newline, in vim-mode it also yanks the newline which makes more sense to me (and `p` sees it as a linewise selection because that's what Utils.coffee does, can't blame it)
 - `y3l` in VIM on the last character will yank the character under the cursor, the newline, and the first character of the next line, and the same in vim-mode

None of the above is affected by this PR or should be. Does this make sense?

spec/operators-spec.coffee
jacekkopecky: Thanks. 8-)

lib/motions/general-motions.coffee
maxbrunsfeld: My general sense is that we should make Atom feel like vim, but not necessarily reimplement all of vim's inconsistencies. The `deleting` option that you added makes complete sense as a fix for this particular bug, but it bothers me that a specific motion like `MoveRight` would have knowledge of a particular type of operator (`deleting`) and take on a different behavior when used with those operators.

Regarding the findings that you shared above, are there any operators whose behavior would *badly* diverge from their behavior in vim if we were to handle `wrapLeftRightMotion` uniformly for all operators (by replacing `if options.deleting` with `if @vimState.mode is 'operator-pending'`)?

jacekkopecky: Ah, well put. I very much agree that *"we should make Atom feel like vim, but not necessarily reimplement all of vim's inconsistencies."* I've switched to checking for `operator-pending` and all my tests still work (as do all the other tests); the findings above change a bit but not in ways that I mind.

lib/vim-state.coffee
```diff
@@ -377,6 +377,9 @@ class VimState
 
     @clearOpStack()
     selection.clear(autoscroll: false) for selection in @editor.getSelections()
+    for cursor in @editor.getCursors()
+      if cursor.isAtEndOfLine() and not cursor.isAtBeginningOfLine()
+        cursor.moveLeft()
```
jacekkopecky: This is so that things like `gUl` don't put the cursor after the end of a line.

lib/motions/general-motions.coffee
```diff
@@ -149,10 +149,16 @@ class MoveLeft extends Motion
 class MoveRight extends Motion
   operatesInclusively: false
 
-  moveCursor: (cursor, count=1) ->
+  moveCursor: (cursor, count=1, options={}) ->
```
maxbrunsfeld: Could you remove the `options` parameter here now?

lib/motions/general-motions.coffee
```diff
@@ -149,10 +149,16 @@ class MoveLeft extends Motion
...
     _.times count, =>
+      wrapToNextLine = settings.wrapLeftRightMotion()
+
+      # when the motion is combined with an operator, we will only wrap to the next line
+      # if we are already at the end of the line (after the last character)
+      wrapToNextLine = false if @vimState.mode is 'operator-pending' and not cursor.isAtEndOfLine()
+
```
maxbrunsfeld: :+1: I like this latest formulation.

maxbrunsfeld:
Great. Thanks for sticking with this through the back and forth.

lib/motions/general-motions.coffee
jacekkopecky: done, thanks for spotting it.

jacekkopecky:
Thanks for patiently explaining your thinking there! And for merging while I'm writing this comment, that was cool. 8-)

PR #556, opened 2015-03-08
:feature: `U` and `u` for upper-casing and lower-casing

jacekkopecky:
The code is adapted from `~` for toggle case. 
I know there is `cmd-K cmd-U` to uppercase in Atom, but it doesn't end the visual mode as VIM would.

### DIFF
--- a/docs/operators.md
+++ b/docs/operators.md
@@ -36,3 +36,7 @@
   * `J` - joins the current line with the immediately following line.
 * [Mark](http://vimhelp.appspot.com/motion.txt.html#m)
   * `m[a-z][A-Z]` - marks the current cursor position
+* [Case](http://vimhelp.appspot.com/motion.txt.html#operator)
+  * `~` or `g~` - toggles case
+  * `gU`, or `U` in visual mode - changes to upper case
+  * `gu`, or `u` in visual mode - changes to lower case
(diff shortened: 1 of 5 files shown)


maxbrunsfeld:
Yeah, I don't think we should need to check the mode in those commands. They can be used as operators (e.g. `gUiw` to uppercase the current word), but that shouldn't require any custom code; it should automatically work the same in visual mode and operator mode, as long as the command modifies the current selection.

maxbrunsfeld:
The reason that `~` contained that `if`/`else` is because `~` isn't an operator when used in normal mode; it just changes the character under the cursor.

maxbrunsfeld:
By the way it's awesome that you're implementing this; thanks.

jacekkopecky:
OK, I'm looking at making `gu`, `gU`, and `g~` work as operators; currently they don't.

jacekkopecky:
Thanks for the thanks, I'm doing it for myself though. 8-)
I think it behaves better now; I wasn't familiar with `g~` and the ilk so I missed that.
And I understand a bit better how vim-mode operators work, that's good too.

maxbrunsfeld:
@jacekkopecky is on :fire:!

PR #560, opened 2015-03-09
:fix: repeating insert operations

### LINKED ISSUE #538: repeat (.) should remember motions
When prefixing several lines I'll often use `I`, `MyPrefix`, `<esc>`, `j` and then `.` to repeat. It currently forgets the `I` (or `A` to suffix) and just inserts the text right there.

If you can point me in the right direction I'll try to contribute a fix when I have some time to work on it.

jacekkopecky:
An `I` or `A` insert operation should be repeated as such, rather than
as a simple `i` or `a` operation.
Fixing this led me to find a bug: repeated insert didn’t position the
cursor correctly at the last inserted character (not after it), so even
repeated `a` or `i` wouldn’t work right; this is fixed too.

Fixes #538 .

### DIFF
--- a/lib/operators/input.coffee
+++ b/lib/operators/input.coffee
@@ -24,6 +24,8 @@ class Insert extends Operator
           @typedText,
           normalizeLineEndings: true
         )
+        cursor = @editor.getLastCursor()
+        cursor.moveLeft() unless cursor.isAtBeginningOfLine()
     else
       @vimState.activateInsertMode()
       @typingCompleted = true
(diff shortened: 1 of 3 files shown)


lib/operators/input.coffee
```diff
@@ -24,6 +24,8 @@ class Insert extends Operator
           @typedText,
           normalizeLineEndings: true
         )
+        cursor = @editor.getLastCursor()
+        cursor.moveLeft() unless cursor.isAtBeginningOfLine()
```
maxbrunsfeld: Can you explain why you added these two lines? I'm thinking we should avoid operating only on the editor's *last* cursor, to make sure that these commands work smoothly with multiple cursors.

maxbrunsfeld:
:zap: This will be a big improvement.

lib/operators/input.coffee
jacekkopecky: This was because line 23 uses getCursorBufferPosition which only deals with the last cursor. If that gets changed, my two lines would want to be updated too. 

(I'm still new here so I go by existing code – I haven't even used multiple cursor much yet, and never in vim-mode.)

maxbrunsfeld: Got it. I didn't realize that repeating insert operations still only worked on the last cursor.

PR #568, opened 2015-03-11
WIP fix #565 : cursor motion in insert mode should affect `.`

### LINKED ISSUE #565: vim inserts are interrupted by cursor movement, vim-mode inserts aren't
In vim, if you move your cursor around (with mouse or with arrow keys), an insert gets confirmed and a new gets started - they can be undone separately, and only the last part can be redone. To demonstrate: 
1. have `1234` in the text editor, with cursor on `2`
2. type `iab<right>cd<esc>` – you will have `1ab2cd34` in the text editor
3. typing `u` should only undo `cd`, typing `.` should only insert `cd` in the current place

But in vim-mode, typing `u` undoes both `ab` and `cd` insertion, and typing `.` inserts `abcd`. 

See [vim docs](http://vimhelp.appspot.com/insert.txt.html#ins-special-special) for more info about this.

I'll now have a quick look if I can fix this easily.

jacekkopecky:
Fixes most of #565 except when I click in the editor, vim-mode doesn't notice. I don't know yet how to fix that, but this PR is better than nothing I guess. 8-)

### DIFF
--- a/keymaps/vim-mode.cson
+++ b/keymaps/vim-mode.cson
@@ -15,6 +15,8 @@
   'ctrl-u': 'editor:delete-to-beginning-of-line'
   'left': 'vim-mode:move-left-insert'
   'right': 'vim-mode:move-right-insert'
+
+'atom-text-editor.vim-mode.insert-mode:not(.autocomplete-active)':
   'up': 'vim-mode:move-up-insert'
   'down': 'vim-mode:move-down-insert'
 
--- a/keymaps/vim-mode.cson
+++ b/keymaps/vim-mode.cson
@@ -13,6 +13,10 @@
 'atom-text-editor.vim-mode.insert-mode':
   'ctrl-w': 'editor:delete-to-beginning-of-word'
   'ctrl-u': 'editor:delete-to-beginning-of-line'
+  'left': 'vim-mode:move-left-insert'
+  'right': 'vim-mode:move-right-insert'
+  'up': 'vim-mode:move-up-insert'
+  'down': 'vim-mode:move-down-insert'
 
 'atom-text-editor.vim-mode:not(.insert-mode)':
   'h': 'vim-mode:move-left'
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -89,7 +89,7 @@ class Motion
     selection.modifySelection => @moveCursor(selection.cursor, count, options)
 
   ensureCursorIsWithinLine: (cursor) ->
-    return if @vimState.mode is 'visual' or not cursor.selection.isEmpty()
+    return if @vimState.mode is 'visual' or @vimState.mode is 'insert' or not cursor.selection.isEmpty()
     {goalColumn} = cursor
     {row, column} = cursor.getBufferPosition()
     lastColumn = cursor.getCurrentLineBufferRange().end.column
@@ -153,9 +153,10 @@ class MoveRight extends Motion
     _.times count, =>
       wrapToNextLine = settings.wrapLeftRightMotion()
 
-      # when the motion is combined with an operator, we will only wrap to the next line
-      # if we are already at the end of the line (after the last character)
-      wrapToNextLine = false if @vimState.mode is 'operator-pending' and not cursor.isAtEndOfLine()
+      # when the motion is in insert mode or is combined with an operator,
+      # we will only wrap to the next line if we are already
+      # at the end of the line (after the last character)
+      wrapToNextLine = false if (@vimState.mode is 'insert' or @vimState.mode is 'operator-pending') and not cursor.isAtEndOfLine()
 
       cursor.moveRight() unless cursor.isAtEndOfLine()
       cursor.moveRight() if wrapToNextLine and cursor.isAtEndOfLine()
--- a/lib/operators/input.coffee
+++ b/lib/operators/input.coffee
@@ -28,6 +28,15 @@ class Insert extends Operator
 
   inputOperator: -> true
 
+# an insert operation following cursor motion in insert mode can be cancelled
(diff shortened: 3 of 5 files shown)


jacekkopecky:
Because my fix generates an InsertCancellable operation with every arrow key motion, the history was growing quite quickly; if InsertCancellable removes itself, it's better after all.

jacekkopecky:
This PR doesn't fix one part of #565: when I click in the editor, vim-mode doesn't notice. I don't know yet how to fix that, but this PR is better than nothing I guess. 8-)

jacekkopecky:
I've added two tests for the new behaviour. Now the PR should be good I hope. 8-)

jacekkopecky:
Rebased to current state of the art, cleaned up.

bronson:
There's a bug with this PR: https://github.com/bronson/vim-mode-next/issues/4

Also, even though it's more Vim-ish, I don't like that right-arrow in insert mode no longer wraps to the next line.

jacekkopecky:
Quite right, sorry about that. Working on a fix.

About the wrapping, @bronson, do I understand correctly that you have wrapLeftRightMotion off, and yet want that motion to wrap in insert mode?

bronson:
Yes, I said that, but I take it back.  Sorry.  :)

That's how my workstation's Vim works -- normal mode doesn't wrap, insert mode does.  But I'm on my laptop now, and neither mode wraps.  Since I haven't noticed the difference until now (we're talking years probably...), I guess I don't care much after all.

I'll get used to whichever way vim-mode settles on.  And I definitely vote to NOT implement whichwrap!

jacekkopecky:
Fixed bronson/vim-mode-next#4 , but need to postpone rebasing to master for now. This PR is related to #609 and should be considered together, and that one needs maintainer guidance.

bronson:
Fix looks great!  Added this PR back to latest vim-mode-next, 0.54.15.

bronson:
This PR breaks autocomplete-plus.

When the completions are shown, I hit down arrow to move to the next completion.  Instead, vim-mode:move-down-insert fires, turns off the autocompletions, and moves one line down in the buffer.

jacekkopecky:
Thank you, good catch. If we find other similar interactions, we can start thinking about how to address this in a less case-by-case way.

jacekkopecky:
Forgot to say explicitly: I added a fix for the autocomplete issue. 8-)

bronson:
Yeah, the `not(.autocomplete-active)` clause might turn into `not(.autocomplete-active|.doc-hints-active|.weatherbug-active|.j-random-plugin|.etc)`.

That said, hopefully there won't be a lot of plugins popping up distractions as I type.  Maybe this is good enough.

jacekkopecky:
Yes @bronson, that's exactly the issue that one immediately imagines. I suspect, though, that due to life being how it is, interactions with *j-random-plugin* would need some other workaround. Notice how only up/down is in `not(.autocomplete-active)` and not left/right; I'm sure *j-random-plugin* will have a different conflicting set. As they say - one is an incident, two is a coincidence, and three's a pattern. I'd like to postpone trying to generalize until I see a pattern. If Atom provides one, I'd love to know about it.

jacekkopecky:
I just realized this approach to fixing #565 isn't right - I've only taken into account arrow-key motions but not any other motions than can happen in insert mode (pgdn, end etc.). I'll redo this PR with `onDidChangeCursorPosition`.

bronson:
Does it make sense to just fix #565 here and do the cursor motion fixes in another PR?

Also, mouse doesn't seem to interrupt insert mode either...  Just wondering if you thought about it and found the fix was too ugly.

jacekkopecky:
@bronson fixing it with `onDidChangeCursorPosition` will be one fix for any motion, including mouse clicks. I have 30e4073c (a biggie) and #767 (a small one) in the queue before I can get to this one.

bronson:
I hear that, first things first.  I was just wondering if that would be an easy way to make this PR smaller / easier to merge?

My random thinking...   I care a lot about fixing giant undos, but not much about forcing the insert cursor to stay on the same line.  (Actually, once that lands, I'm going to figure out a way to turn it off.  :)  Turns out I really do like insert mode feeling just like Atom.  I know, I'm waffling like crazy on this one!)

jacekkopecky:
@bronson I'm not sure what you mean about cursor positioning there

bronson:
Sorry, let me back up...   This PR does two things, right?
- ensures arrow key motion (and hopefully mouse clicks) interrupt insert mode keeping undos small
- restricts the cursor to the current line using keymaps now, and hopefully onDidChangeCursor position in the future

Just curious if it would help mergeability to split these two behaviors into separate PRs.  (ordered of course -- one would have to be applied before the other)

jacekkopecky:
Ah I see, I'd forgotten about restricting left/right motions in case wrapLeftRightMotion is off. 
I've been trying to do the interrupting of undo with `onDidChangeCursorPosition` but it's a huge pain - of course Atom will report a cursor position change on every input (at least it says the cursor motion was caused by text change, except backspace *doesn't* report the motion was caused by text change); and bracket-matcher's insertion of `()` when the user types `(` also causes a cursor motion. 

At the moment I'm thinking that some departure from VIM is warranted here - perhaps insert undo should work like normal, non-vim-mode, typing undo? 

My littlest one has been keeping me busy these days but soon I should be able to make an alternate PR to show what I mean.

bronson:
Sure, vim-mode undo working just like Atom undo sounds fine to me.  Atom's undo grouping used to be atrocious but seems pretty good now.  As long as the entire insert session doesn't get grouped into a single undo, I'd probably be happy with just about anything.

lee-dohm:

PR #570, opened 2015-03-11
vim-mode should dispose all its subscriptions

jacekkopecky:
If I understand it correctly, this patch should be appropriate. If not, at least I'll improve my understanding. 8-)


maxbrunsfeld:
This is right.

PR #573, opened 2015-03-11
add replace mode

jacekkopecky:
This addresses #562; also it partly addresses #564.

I got the inspiration for this implementation from @brunetton's `atom-overtype-mode` package.

*(edited)*
**Features:**
 * `backspace` acts as undo
 * block cursor and status bar indication of replace mode

**Limitations:**
 * `.` with multiple cursors may eat a character for one of those cursors, but I’ve no idea why

PR #574, opened 2015-03-12
:feature: show submode in status bar

jacekkopecky:
This is a suggestion that we may want to show the submode in the status bar; the PR already includes values for replace mode, proposed in #573.

### DIFF
--- a/lib/status-bar-manager.coffee
+++ b/lib/status-bar-manager.coffee
@@ -1,9 +1,14 @@
 {Disposable, CompositeDisposable} = require 'event-kit'
 
+# mode names can be combined with submode names with the `$` character
+# the order matters - the last matching entry wins
 ContentsByMode =
   insert:  ["status-bar-vim-mode-insert", "Insert"]
   command: ["status-bar-vim-mode-command", "Command"]
   visual:  ["status-bar-vim-mode-visual", "Visual"]
+  visual$characterwise: ["status-bar-vim-mode-visual", "Visual Char"]
+  visual$linewise:      ["status-bar-vim-mode-visual", "Visual Line"]
+  visual$blockwise:     ["status-bar-vim-mode-visual", "Visual Block"]
 
 module.exports =
 class StatusBarManager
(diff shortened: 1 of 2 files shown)


lib/status-bar-manager.coffee
```diff
@@ -14,9 +19,10 @@ class StatusBarManager
 
   initialize: (@statusBar) ->
 
-  update: (currentMode) ->
+  update: (currentMode, currentSubmode) ->
+    currentFullMode = currentMode + "$" + currentSubmode if currentSubmode?
```
maxbrunsfeld: I like it; nice solution. Nitpicks:

* I'd rather use `.` as the delimiter than `$`. It would just mean we'd have to quote the keys.
* I think you could avoid depending on the order of `ContentsByMode` if you did this:

```coffee
currentMode = currentMode + "." + currentSubmode if currentSubmode?
for mode, [klass, html] of ContentsByMode
  if mode is currentMode
    # ...  
```
```

jacekkopecky: Thank you.
I've committed a fix in 91069db61fd16d6f7c6a03880cceea0fc8eeb235 so the code doesn't rely on order, keeping the ability to match only on the `mode`. If we don't keep `currentFullMode` we'd have to enumerate all submodes in ContentsByMode.
Which alternative would you prefer?

jacekkopecky:
Oh, rebasing the feature branch and forcing the push drops commit line comments - sorry @maxbrunsfeld. Is there a better way to get feature branches in sync with master?

lib/status-bar-manager.coffee
maxbrunsfeld: Oh, I see what you're saying. I think the most straightforward thing to do is to use the code snippet I wrote above; we just need to enumerate *all* possible combinations of `mode` and `submode` in `ContentsByMode`, rather than abbreviating `visual.characterwise` as `visual`.

I don't think we can use a `break` statement like you have on line 30, because it could incorrectly prevent classes from being removed from the `@element` div.

jacekkopecky: You're right, I missed that. OK, enumeration it is; the longer I've thought about it the better this option has sounded anyway. Commit coming soon.

jacekkopecky:
And I still can't get it as simple as you suggested, @maxbrunsfeld : 
Since multiple entries in `ContentsByMode` share the same class name, there was a risk that a later entry with the same class name but not a matching mode would remove the class set by a preceding entry with a matching mode. In other words, in visual.characterwise, the `@element` would not have the class `status-bar-vim-mode-visual` because visual.blockwise would remove it.
So 5490b33 does things differently still. Comments welcome.

jacekkopecky:
Thanks! 8-)

PR #581, opened 2015-03-14
Ensure delete-word tries to stop at the end of the line

### LINKED ISSUE #550: Delete word (dw) should not delete line break
If I have a line of code and I try to delete the last word, `dw` joins the lower line to the current line.  I would expect more of the behavior that `de` provides.  It deletes the word, but doesn't join the lower line.

```
test test
nope
```

becomes

```
test nope
```

I would expect

```
test
nope
```

dsandstrom:
Fixes #550.
- I had to add a different word regex to add `$`.  I'd rather have an option built into the Cursor class instead of adding it to this package.  I can submit that PR if that is a better approach.
- I wanted to set the `allowEOL` option from _vim-state.coffee_, but couldn't figure out how to do that.  Right now, the custom word regex is set for all delete operators.
- In vim, I noticed that `dw` also deletes the preceeding space, which this PR doesn't fix.

### DIFF
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -201,7 +201,18 @@ class MoveToNextWord extends Motion
   wordRegex: null
   operatesInclusively: false
 
+  wordRegExp: (cursor, {includeNonWordCharacters}={}) ->
+    includeNonWordCharacters ?= true
+    nonWordCharacters = atom.config.get('editor.nonWordCharacters', scope: cursor.getScopeDescriptor())
+    segments = ["^[\t ]*$"]
+    segments.push("[^\\s#{_.escapeRegExp(nonWordCharacters)}]+")
+    segments.push("$")
+    if includeNonWordCharacters
+      segments.push("[#{_.escapeRegExp(nonWordCharacters)}]+")
+    new RegExp(segments.join("|"), "g")
+
   moveCursor: (cursor, count=1, options) ->
+    @wordRegex = @wordRegExp(cursor)
     _.times count, =>
       current = cursor.getBufferPosition()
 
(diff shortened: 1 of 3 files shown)


jacekkopecky:
@dsandstrom can you please clarify what you mean by `dw` deleting the preceding space in VIM? I can't seem to see such behaviour. I'm asking because this is related to #555 and related stuff I've had my eye on. 8-)

dsandstrom:
I guess I was mistaken.  I was thinking that if you entered `dw` when your cursor was before the second word (`first |second`), that vim was also deleting the space after the first word.  But it is not the case.

chrisvdb:
hi guys, is this fix production ready? If so, can this fix be included in the next release? Issue #550 hits me 10+ times per day.

dsandstrom:
@chrisvdb No, it's more of a proof of concept.  I'm waiting to hear from @maxbrunsfeld or @kevinsawicki for some feedback.

Zaba999:
:+1:

lib/motions/general-motions.coffee
```diff
@@ -201,7 +201,18 @@ class MoveToNextWord extends Motion
   wordRegex: null
   operatesInclusively: false
 
+  wordRegExp: (cursor, {includeNonWordCharacters}={}) ->
```
maxbrunsfeld: Is this method ever called with a second argument? I only see the one call on line 215.

dsandstrom: No, I just copied cursor's regex function: https://github.com/atom/atom/blob/37cb6b6d9e2be35af51b4e7e812b833efc24ce11/src/cursor.coffee#L642.  My original intent was to merge the changes back to Atom.

lee-dohm:

PR #582, opened 2015-03-14
Use TextEditorElement for VimCommandModeInputView

tmm1:
fixes a bug where you're unable to see the text you're typing when hitting `/` to search

searching is working on this PR, but I introduced a regression into Motion.Till where the search input box shows up whereas previously it was invisible

cc @maxbrunsfeld

### DIFF
--- a/lib/view-models/vim-command-mode-input-view.coffee
+++ b/lib/view-models/vim-command-mode-input-view.coffee
@@ -4,15 +4,20 @@ module.exports =
 class VimCommandModeInputView extends View
   @content: ->
     @div class: 'command-mode-input', =>
-      @div class: 'editor-container', outlet: 'editorContainer', =>
-        @subview 'editor', new TextEditorView(mini: true)
+      @div class: 'editor-container', outlet: 'editorContainer'
 
   initialize: (@viewModel, opts = {})->
     if opts.class?
       @editorContainer.addClass opts.class
 
     if opts.hidden
+      # TODO: this no longer works with atom-text-editor
       @editorContainer.addClass 'hidden-input'
+
+    @editorElement = document.createElement "atom-text-editor"
+    @editorElement.classList.add('editor')
+    @editorElement.getModel().setMini(true)
+    @editorContainer.append(@editorElement)
 
     @singleChar = opts.singleChar
     @defaultText = opts.defaultText ? ''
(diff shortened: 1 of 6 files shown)


lib/view-models/vim-command-mode-input-view.coffee
```diff
@@ -4,16 +4,21 @@ module.exports =
...
-        @subview 'editor', new TextEditorView(mini: true)
+      @div class: 'editor-container', outlet: 'editorContainer'
 
   initialize: (@viewModel, opts = {})->
     if opts.class?
       @editorContainer.addClass opts.class
 
     if opts.hidden
+      # TODO: this no longer works with atom-text-editor
       @editorContainer.addClass 'hidden-input'
```
tmm1: The `hidden-input` class was a special case in the deprecated TextEditorView. I'm not sure what the equivalent is for TextEditorElement

MattKunze:
fwiw I have this updated in #545

maxbrunsfeld:
@MattKunze the nice thing about this approach is that it uses the newer `TextEditorElement` class, which avoids having to use `atom-space-pen-views`. That module will work, but it depends on `space-pen` and jQuery, so it'd be nice to use `TextEditorElement` directly instead. It just means we need to do a little more work to update the tests to not use the jQuery interface.

@tmm1 did you still want to work on this? If not, I can pick it up here. I don't think the `hidden-input` thing needs to change. That class is being applied to a `div` that *contains* the text editor element, not the element itself, so I would think it would continue to work.

MattKunze:
It didn't seem like the optimal approach, I was just following the bouncing balls of documentation regarding upgrading the internal views.

tmm1:
Feel free to pick this up. I noticed the old deprecated class had some special handling for hidden-input, and I wasn't able to make the new view hidden.

MattKunze:
I think the motion regression is with handling single-character inputs. The input used to show up, but it was closed after the first character that was typed. That's what hung me up on the other PR was how to handle that with the updated internals.

tmm1:
> The input used to show up, but it was closed after the first character that was typed.

In my testing it never used to show up. There's code in the plugin to explicitly hide it in the singleChar case.

On this PR, it does show up and it goes away as soon as you type a character. But it's awkward to see it pop up briefly and definitely distracting.

maxbrunsfeld:
Fixes #544

lib/view-models/vim-command-mode-input-view.coffee
```diff
@@ -24,35 +29,29 @@ class VimCommandModeInputView extends View
...
-  stopHandlingEvents: ->
-    if @singleChar?
-      @editor.find('input').off 'textInput', @autosubmit
-    @editor.off 'core:confirm', @confirm
-    @editor.off 'core:cancel', @cancel
-    @editor.find('input').off 'blur', @cancel
-
-  autosubmit: (event) =>
-    @editor.setText(event.originalEvent.data)
+      @editorElement.addEventListener('textInput', @autoSubmit)
```
tmm1: This doesn't seem to work with addEventListener. Maybe the event needs to be keyup or something different

maxbrunsfeld: Yeah, I can't remember why that doesn't work. Observing text changes on the model seems to work ok though.

tmm1:
Thanks @maxbrunsfeld, this is working great now!

PR #590, opened 2015-03-28
fix cursor position after operator

jacekkopecky:
`gU$` and such could put the cursor at the end of the line, not the last character as expected in vim mode

### DIFF
--- a/spec/operators-spec.coffee
+++ b/spec/operators-spec.coffee
@@ -1364,6 +1364,13 @@ describe "Operators", ->
       keydown("e")
       expect(editor.getText()).toBe 'ABC\nXyZ'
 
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
I saw some similar code to this in #555. Does that cover this case too, or is this one different?

jacekkopecky:
right, it may be obsolete, I'll check. If so, should I keep the PR with just the specs, or should we just drop it?

maxbrunsfeld:
> If so, should I keep the PR with just the specs

Good idea.

jacekkopecky:
Done, indeed the change in #555 did fix this.

maxbrunsfeld:
Great.

PR #558, opened 2015-03-09
add `z s` and `z e` horizontal scrolling

jacekkopecky:
This will scroll the window horizontally so that the cursor is at (or near) the left edge of the screen for `z s`, or near the right edge of the screen for `z e`.
Sorry but I've no idea how to test for this behaviour in the specs. Please advise.

PR #591, opened 2015-03-28
fix: yank should not scroll

jacekkopecky:
Yank currently shows the target of the motion, for example on a line that's longer than the screen, `y$` will scroll so that the end of the line is visible, and `yn` will scroll to the next search match, even if that means that the cursor goes off-screen. This PR fixes that.
I don't know how to spec it, similarly to #558 .

### DIFF
--- a/spec/operators-spec.coffee
+++ b/spec/operators-spec.coffee
@@ -100,7 +100,6 @@ describe "Operators", ->
           keydown('u')
           expect(editor.getText()).toBe "abc\n012345\n\nxyz"
 
-
       describe "with vim-mode.wrapLeftRightMotion", ->
         beforeEach ->
           editor.setText("abc\n012345\n\nxyz")
@@ -160,7 +159,6 @@ describe "Operators", ->
           expect(editor.getCursorScreenPosition()).toEqual [0, 1]
           expect(vimState.getRegister('"').text).toBe '0123\n\nx'
 
-
     describe "on an empty line", ->
       beforeEach ->
         editor.setText("abc\n012345\n\nxyz")
@@ -177,7 +175,6 @@ describe "Operators", ->
         keydown('x')
         expect(editor.getText()).toBe "abc\n012345\nxyz"
         expect(editor.getCursorScreenPosition()).toEqual [2, 0]
-
 
   describe "the X keybinding", ->
     describe "on a line with content", ->
@@ -206,7 +203,6 @@ describe "Operators", ->
         expect(editor.getText()).toBe 'ab2345'
         expect(editor.getCursorScreenPosition()).toEqual [0, 2]
         expect(vimState.getRegister('"').text).toBe '\n'
-
 
     describe "on an empty line", ->
       beforeEach ->
@@ -939,7 +935,7 @@ describe "Operators", ->
       describe "yanking many lines forward", ->
         it "does not scroll the window", ->
           editor.setCursorBufferPosition [40, 1]
-          top40 = editor.getScrollTop()
+          previousScrollTop = editor.getScrollTop()
 
           # yank many lines
           keydown('y')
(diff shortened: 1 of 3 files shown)


jacekkopecky:
Just found #417 which led me to find a bug in my PR - a yank backwards now doesn't scroll to the newly positioned cursor (because a yank backwards moves the cursor - why should a yank backwards move the cursor when a yank forward doesn't?)
But interestingly `yk` doesn't move the cursor in vim-mode but does in VIM.

jacekkopecky:
Hmmm, can't seem to be able to replicate the scrolling bug I was talking about, maybe df13c6b50abc73972da455e37f280bd82c4bb9a6 fixed it.

jacekkopecky:
this one no longer works in the newest atom, I'm looking into fixing it

jacekkopecky:
all new and improved:
This PR makes the following improvements to Yank:
- cursor moves to the beginning of the yanked range
- but if the motion was linewise, it stays on the same column as it was, so it only moves to the first yanked line
- cursor doesn't move for forwards yanks
- screen doesn't move if the cursor doesn't
- all properly tested

Without this PR, `yG` scrolls the screen so the cursor is at the top (unless the end of the file is visible when doing the `yG`) because the `G` motion actually moves the cursor to the end of the file, then the cursor position is restored and the screen then autoscrolls from the bottom of the file to follow the cursor.

jacekkopecky:
@maxbrunsfeld any thoughts on this new and improved version, please?

spec/scroll-spec.coffee
```diff
@@ -18,6 +18,39 @@ describe "Scrolling", ->
...
+        editor.setCursorBufferPosition [161, 0]
+        top161 = editor.getScrollTop()
+        expect(top161-top100).toEqual(610)
+
+        editor.setCursorBufferPosition [0, 0]
+        expect(editor.getScrollTop()).toNotBe top101
+        editor.setCursorBufferPosition [101, 0]
+        expect(editor.getScrollTop()).toBe top101
+
+
```
maxbrunsfeld: What is this test for exactly? It seems like this behavior is covered in Atom's core test suite.

spec/operators-spec.coffee
```diff
@@ -913,6 +926,48 @@ describe "Operators", ->
...
+        editor.setDefaultCharWidth(10)
+        text = ""
+        for i in [1..200]
+          text += "#{i}\n"
+        editor.setText(text)
+
+      describe "yanking many lines forward", ->
+        it "does not scroll the window", ->
+          editor.setCursorBufferPosition [40, 1]
+          top40 = editor.getScrollTop()
```
maxbrunsfeld: A minor aesthetic point: I think it would fit in a bit better w/ our test suite if you renamed `top40` to `previousScrollTop` (and similarly for `top140` below).

spec/operators-spec.coffee
```diff
@@ -913,6 +926,48 @@ describe "Operators", ->
...
+          keydown('y')
+          keydown('6')
+          keydown('0')
+          keydown('G', shift: true)
+
+          expect(editor.getScrollTop()).toNotEqual top140
+          expect(editor.getCursorBufferPosition()).toEqual [59, 1]
+          expect(vimState.getRegister('"').text.split('\n').length).toBe 83
+
+
```
maxbrunsfeld: Could you remove this second blank line?

maxbrunsfeld:
Other than some minor comments, looks great!

spec/scroll-spec.coffee
jacekkopecky: I guess this was a way for me to confirm my expectations; I'll remove it.

jacekkopecky:
Thank you, all fixed.

spec/operators-spec.coffee
jacekkopecky: I removed other double blank lines too, is that desirable?

PR #606, opened 2015-04-05
`%` should not affect `n`

jacekkopecky:
the `%` motion in VIM does not become the motion repeated with `n`;
this PR fixes this in vim-mode
Let me know if this way of adding exceptions is not the preferred one, and if so please point me to the preferred one. 8-)

this also removes some unnecessary repetition in search motion constructors


jacekkopecky:
actually, let's hold this one for a few moments please – I may be making this PR obsolete with the next one, it seems

Issue #578, opened 2015-03-13
// matches literal /


jazzpi:
According to http://pubs.opengroup.org/onlinepubs/9699919799/utilities/ex.html#tag_20_40_13_58, the regular expression // should be equivalent to the last regular expression encountered. // matches the next / however.

Vim also shows this behavior for ??

jazzpi:
I think this could be combined with #458 since that bascially highlights the last regular expression.

PR #607, opened 2015-04-05
Global current search

### LINKED ISSUE #190: Search repeat no longer works across buffers
This broke somehow with #183. I played with it a little bit, but the refactoring required was actually kind of heavy, so I want to let #184 land before I start messing around.

jacekkopecky:
Make current search part of the global state.
Also prevents `%` from affecting `n`; and also puts the word searched for by `*` and `#` into search history.
Obsoletes #606. Fixes #190 I think.

### DIFF
--- a/lib/view-models/search-view-model.coffee
+++ b/lib/view-models/search-view-model.coffee
@@ -10,7 +10,7 @@ class SearchViewModel extends ViewModel
     atom.commands.add(@view.editorElement, 'core:move-down', @decreaseHistorySearch)
 
   restoreHistory: (index) ->
-    @view.editorElement.getModel().setText(@history(index).value)
+    @view.editorElement.getModel().setText(@history(index))
 
   history: (index) ->
     @vimState.getSearchHistoryItem(index)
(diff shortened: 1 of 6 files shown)


lib/view-models/search-view-model.coffee
```diff
@@ -30,5 +30,5 @@ class SearchViewModel extends ViewModel
       @restoreHistory(@historyIndex)
 
   confirm: (view) =>
-    @vimState.pushSearchHistory(@)
```
jazzpi: If we add something like

```coffeescript
if @view.value is '/'
  # somehow repeat current search without circular referencing
```

could this fix #578?

jacekkopecky: Yes, I think so: `//<enter>`, `??<enter>`, `/<enter>` and `?<enter>` would all be easy to implement with this, I think. 
I'll be happy to try and add that if the powers that be (@maxbrunsfeld?) like this PR.

maxbrunsfeld: Yeah, I'll definitely merge this. Do you mind if I do so now, and you can do the features you mentioned above on a separate PR?

jacekkopecky: Oh go ahead, of course. 8-)

PR #609, opened 2015-04-06
Insert with count, repeating `.` with prefix

### LINKED ISSUE #580: Does not repeat inserts
Pressing `30i`, typing some characters, and then pressing `[esc]` should repeat that insert operation the expressed (in this case 30) number of times.

For example `3i#[esc]` should print `###`

Although other repeat operations work properly, this does not seem to work.

jacekkopecky:
Hi, this one has two extra things: it started as insert with count, then I realized I really wanted it over #565, and also I really needed to fix handling of `.` with count.
So here's what this PR does:
- Repeated insert such as `3i` and `3a` now works (fixes #580)
- Also, `.` with a count (such as `3.`) now changes the old prefix of any command:
  - for example `2dd3.` will delete 2 and then 3 lines, a subsequent `.` will again delete 3 lines;
  - similarly `dl3.` will delete a character and then 3 more, a subsequent `.` will again delete 3 characters.
- Includes #568 .

If you want this PR split, please say which part should go first.

### DIFF
--- a/keymaps/vim-mode.cson
+++ b/keymaps/vim-mode.cson
@@ -15,6 +15,8 @@
   'ctrl-u': 'editor:delete-to-beginning-of-line'
   'left': 'vim-mode:move-left-insert'
   'right': 'vim-mode:move-right-insert'
+
+'atom-text-editor.vim-mode.insert-mode:not(.autocomplete-active)':
   'up': 'vim-mode:move-up-insert'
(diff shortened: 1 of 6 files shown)


jacekkopecky:
#568 is WIP again so this PR will have to be delayed.

jacekkopecky:
I can get back to this one when we're done with #849 or #568.

lee-dohm:

PR #613, opened 2015-04-08
Add coffeelint support

mnquintana:
Refs atom/atom#6230, atom/atom#5982


kevinsawicki:
Nice, this caught some good stuff :+1:

Issue #614, opened 2015-04-09
Ctrl+F or Ctrl+B reset visual line mode selection


Valve:
If I select multiple lines in visual line mode (shift+v, j,j,j), and then press Ctrl+F to extend the selection to the next screen, the visual line selection disappears.

Expected behavior:

Visual line mode should select the next screen, forwarded by Ctrl+F

daniloisr:
It's similar to this issue https://github.com/atom/vim-mode/issues/593

Valve:
Since #616 got merged, will this have a fix as well?

maxbrunsfeld:
I think this has been fixed for some time. Can anyone reproduce this with Atom 1.1.0 and vim-mode 0.63?

morenoh149:
dont' think this works

maxbrunsfeld:
@morenoh149 I think this has been fixed for a few releases; I actually meant to close this. Are you on the latest Atom and vim-mode?

morenoh149:
I just updated to 0.63. Maybe Im confused? I came here searching for why `ctrl-f` moves the screen down but not my cursor.

So I do `ctrl-f` then when I do `j` my screen jumps back to where the cursor stayed.

maxbrunsfeld:
I remember a bug like that being fixed... what version of Atom are you on?

morenoh149:
oh it's fixed now. I hadn't restarted.

maxbrunsfeld:
:+1: Thanks for the update.

Issue #593, opened 2015-03-29
C-d / C-u in visual mode


josephcc:
In vim, you can use C-d / C-u to expand your selection while in visual mode, but in atom/vim-mode it would simply put you back in the command mode.

daniloisr:
I'm woking with it, and my idea is to switch the C-d behaviour from a `Scroll` subclass to a `MoveToLine` subclass. I write a sketch for it and apparently it's working https://github.com/daniloisr/vim-mode/commit/544f5dbd8e0e86d9fd0097a830624aebbb6ca88e.

What do you guys think? Can I continue with this approach?
Thx

maxbrunsfeld:
@daniloisr Yeah, your approach is good. I would love to merge a PR fixing this. Are you interested in fixing #614 too?

daniloisr:
@maxbrunsfeld Nice! I will prepare a PR for both issues soon. Thx

bronson:
This can be closed?  It appears to work in 0.200.0 / 0.48.0.

daniloisr:
@bronson I think it be closed

maxbrunsfeld:
Thanks!

PR #616, opened 2015-04-09
Keep visual mode for some motions

daniloisr:
Keep visual mode for the following motions:
- ctrl-f
- ctrl-b
- ctrl-d
- ctrl-u

Solves #593 and #614


daniloisr:
There is some other way to get the editor height without this code?

```coffee
firstScreenRow = @editor.getFirstVisibleScreenRow()
lastScreenRow = @editor.getLastVisibleScreenRow()
height = lastScreenRow - firstScreenRow
```

daniloisr:
I think that it's ok now :v:

jacekkopecky:
@daniloisr this PR changes the behavior of these commands - in VIM they scroll the screen so the cursor stays on the same line on the screen, here they don't. I think we'd want to keep the scrolling behavior as well as fixing how these motions don't affect the visual mode. (Yes, they do seem to be motions rather than scrolls.)

daniloisr:
@jacekkopecky you are right friend, sorry about that, I have used ctrl-d and ctrl-u for a long time thinking that they are motions, but I read the docs right now and saw that they are scrolls.

I could try to fix they as scrolls, but for now, what you think about I letting this PR only for ctrl-f and ctrl-b fix?

daniloisr:
Ops, ctrl-f and ctrl-b are scroll too, :/

I already started working on that, soon I will fix this PR

daniloisr:
@jacekkopecky I have changed the code and added visual selection to `Scroll` class*, and before I move to `ctrl-f` and `ctrl-b` implementation I would like to know your opinion, did you think that this way is correct?

\* I used the `select` method from `Motion` class, so maybe its better to join these codes... I'm studying it now

jacekkopecky:
Hi @daniloisr, this looks interesting, I'm hoping to get enough time to focus on this in a few days, hope that's OK.

daniloisr:
Finished this PR.

I think that after this PR I will work on other related problems like keeping selection when scrolling the whole screen with `ctrl-y`, because today we lose the selection.

maxbrunsfeld:
@daniloisr Thanks for working on this! Because these commands move the cursor and modify the selection, I would really prefer to implement them as subclasses of `Motion`. That way we wouldn't have to duplicate the logic for updating the selection.

You should be able to implement `Motion::moveCursor(cursor)` so that it positions the cursor correctly using `Cursor::setBufferPosition` or `::setScreenPosition`, and then override `Motion::select` and `Motion::execute` so that they update the `@editor`'s scroll position at the end.

Let me know if that approach makes sense to you, I can give you more detailed guidance if it'd be helpful.

spec/scroll-spec.coffee
```diff
@@ -121,31 +121,79 @@ describe "Scrolling", ->
...
+        vimState.activateVisualMode()
+        keydown('d', ctrl: true)
+        expect(editor.getSelectedText()).toEqual [42..52].join("\n").slice(1)
+
+      it "selects on linewise mode", ->
+        vimState.activateVisualMode('linewise')
+        keydown('d', ctrl: true)
+        expect(editor.getSelectedText()).toEqual [42..52].join("\n").concat("\n")
+
+    fdescribe "the ctrl-f keybinding", ->
```
maxbrunsfeld: Same here.

spec/scroll-spec.coffee
```diff
@@ -121,31 +121,79 @@ describe "Scrolling", ->
...
+        vimState.activateVisualMode()
+        keydown('u', ctrl: true)
+        expect(editor.getSelectedText()).toEqual [32..42].join("\n").slice(1)
 
+      it "selects on linewise mode", ->
+        vimState.activateVisualMode('linewise')
+        keydown('u', ctrl: true)
+        expect(editor.getSelectedText()).toEqual [32..42].join("\n").concat("\n")
+
+    fdescribe "the ctrl-b keybinding", ->
```
maxbrunsfeld: Gotta remove the `f` here before we merge, too.

daniloisr:
Hi @maxbrunsfeld. My first approach was to create a `Motion` subclass, but these commands also scrolls the screen (like @jacekkopecky said) and we have the [`Scroll` class that defines some useful stuffs like `@scrolloff` and `@rows`](https://github.com/atom/vim-mode/blob/master/lib/scroll.coffee#L5), so if I create a `Motion` subclass I will need them too.

I was thinking a lot about it, and my solution was extract the selection logic to a mixin because I think that `Motion` classes can use this mixin later (I can implement it next).

Do you think that implementing a `Motion` subclasse still a better solution? :sweat_smile:

maxbrunsfeld:
Do you still have your commits where you wrote it as a `Motion` subclass? I don't see them on this branch. You could just push them to a different branch if so; I'm curious what it looked like.

> we have the Scroll class that defines some useful stuffs like @scrolloff and @rows
That's true. I just think the `Motion` class defines a lot *more* useful stuff than the `Scroll` class, and I think `Motion` is a more correct way to think of these commands, since they can be composed with operators.

daniloisr:
@maxbrunsfeld sorry, I deleted the branch =P. But dont worry, the solution was almost like [the class `MoveToBottomOfScreen`](MoveToBottomOfScreen), I just did the right math, when @jacekkopecky commented I had not implemented the scrolling yet.

I liked your idea, soon I can write a new code doing the scroll in a `Motion` subclass. What you think?

maxbrunsfeld:
:+1: Sounds good. Sorry for not giving you feedback earlier. I definitely want to merge this fix.

daniloisr:
@maxbrunsfeld done! Later I will try to improve the way the selection is made, I have some felling that it may be simpler.

The build has failed due module download problem, I tried to rerun the build but I dont know how.

maxbrunsfeld:
Looks *really* nice. Re-running the build now.

lib/motions/general-motions.coffee
```diff
@@ -379,11 +379,56 @@ class MoveToMiddleOfScreen extends MoveToScreenLine
...
+class ScrollKeepingCursor extends MoveToLine
+  previousFirstScreenRow: 0
+  currentFirstScreenRow: 0
+
+  constructor: (@editor, @vimState) ->
+    super(@editor, @vimState)
+
+  select: (count, options) ->
+    @scrollScreen(count)
+    super(count, options)
```
maxbrunsfeld: So I just tried out your changes. I noticed that when in visual mode, the cursor's position is not preserved relative to the screen. I think you may need to call `@scrollScreen()` *after* calling `super` here:

```coffee
result = super
@scrollScreen(count)
result
```

lib/motions/general-motions.coffee
```diff
@@ -379,11 +379,56 @@ class MoveToMiddleOfScreen extends MoveToScreenLine
...
+
+  execute: (count) ->
+    @scrollScreen(count)
+    super(count)
```
maxbrunsfeld: You may need to do the same thing here, but I'm not sure.

lib/motions/general-motions.coffee
daniloisr: @maxbrunsfeld fixed it storing the `destination` of  `setScrollTop` before calling `super` and setting the `@editor.setScrollTop` again. I know that this is not the best solution but this happen because the `Motion.select` is changing the scroll value, I tried to sove this changing the `moveSelectionLinewise` and `moveSelectionInclusively` code but I could find a better solution.

daniloisr:
@maxbrunsfeld I think it's ok now o/

lib/motions/general-motions.coffee
```diff
@@ -379,11 +379,62 @@ class MoveToMiddleOfScreen extends MoveToScreenLine
     height = lastScreenRow - firstScreenRow
     Math.floor(firstScreenRow + (height / 2))
 
+class ScrollKeepingCursor extends MoveToLine
+  previousFirstScreenRow: 0
+  currentFirstScreenRow: 0
+
+  constructor: (@editor, @vimState) ->
+    super(@editor, @vimState)
```
maxbrunsfeld: You can remove this `constructor` definition, since it only calls `super`.

lib/motions/general-motions.coffee
```diff
@@ -379,11 +379,62 @@ class MoveToMiddleOfScreen extends MoveToScreenLine
...
+
+  select: (count, options) ->
+    final_destination = @scrollScreen(count)
+    super(count, options)
+    @editor.setScrollTop(final_destination)
+
+  execute: (count) ->
+    final_destination = @scrollScreen(count)
```
maxbrunsfeld: We usually don't use `snake_case` variable names. I would say call this variable `scrollDestination`.

maxbrunsfeld:
Looks great! Other than some minor comments, I'll merge.

daniloisr:
@maxbrunsfeld thats it, thx for the review =)

PR #621, opened 2015-04-12
Fix delete with empty selections

jacekkopecky:
With multiple cursors, if a motion selects nothing and other motions select something, delete operations would correctly delete the selections, but the empty selections would delete a character, which is unexpected. 
For example, with two cursors, `dtx` should delete until 'x' only on the line that has an 'x' - of only one does, the other line would delete a character; if neither line has x, nothing would be deleted. 
Another example: `d0` with two cursors, one at the beginning of the first line and another on the second line but not its beginning, should not delete the first character of the line.

### DIFF
--- a/spec/operators-spec.coffee
+++ b/spec/operators-spec.coffee
@@ -390,7 +390,7 @@ describe "Operators", ->
 
     describe "with multiple cursors", ->
       it "deletes each selection", ->
-        editor.setText("abcd\n1234\nABCD")
+        editor.setText("abcd\n1234\nABCD\n")
         editor.setCursorBufferPosition([0, 1])
         editor.addCursorAtBufferPosition([1, 2])
         editor.addCursorAtBufferPosition([2, 3])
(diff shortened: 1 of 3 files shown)


spec/operators-spec.coffee
```diff
@@ -390,7 +390,7 @@ describe "Operators", ->
 
     describe "with multiple cursors", ->
       it "deletes each selection", ->
-        editor.setText("abcd\n1234\nABCD")
+        editor.setText("abcd\n1234\nABCD\n")
```
jacekkopecky: the missing `\n` above was triggering another bug uncovered by this PR's fix: bug #622 is quite unrelated to this PR but the bug fixed by this PR was behaving just like we expect `de` to behave at the end of the file. With the `\n` above, the test still tests the same expected behaviour, but without triggering #622.

maxbrunsfeld:
Great catch. Thanks!

Issue #596, opened 2015-03-30
`/` may be better handled by ex-mode


lloeki:
`/` is in fact a shortcut to `:/`. From vim's `:help :`:

> Command-line mode is used to enter Ex commands (":"), search patterns ("/" and "?"), and filter commands ("!").

Besides, this may help vim-mode get rid of some code such as the search view, which is duplicated (probably due to aforementioned design) between vim-mode and ex-mode.

This may help fix issues such as #578.

jazzpi:
I'm not quite sure about this. Since <kbd>/</kbd> is closely related to <kbd>n</kbd>, <kbd>N</kbd>, <kbd>*</kbd> and <kbd>#</kbd>, those should then probably also be handled by ex-mode as well as potential highlighting...
See also #458 and #493.
As well as that, forcing users to install `ex-mode` as a whole just to be able to use <kbd>/</kbd> seems a bit overkill.

dimpase:
Historically, vi(m) is an add-on of ex, and as such it's only logical that ex-mode and vim-mode should be basically one mode. Vim-mode without ex-mode feels rather feature-poor for any advanced vi(m) user.

jazzpi:
Sure, but a lot of people are apparently fine with vim-mode without ex-mode (vim-mode has about 40x as many downloads as ex-mode). See #50 about the reasoning behind splitting ex into a different package. Following that philosophy, I don't think putting <kbd>/</kbd> into ex-mode is a good idea, although it would simplify some things.

dimpase:
arguments in #50 are just childish; after 20+ years of using vi(m) I don't want to waste time learning yet another way to do complicated things like formatting paragraphs and complicated regex-based replacements, really, and surely I'm not alone.

And measuring success by number of downloads makes little sense, either. You know, pr0n downloads surely beat Atom downloads by many orders of magnitude. ;-)

nnjpp:
I view ex-mode's download count as a sign of a problem. I suspect that it has less downloads because not enough people are willing to dig into github issues to find out that it even exists. I would bet that abandonment comes first in many cases. I'm not buying the notion that only 2.5% of vim users value ex commands.

jacekkopecky:
This is a great discussion to have.

Ex is old and powerful. I don't know that anybody has done an ergonomic comparison between ex-mode and atom ways of doing things that don't fit within the few keystrokes of vim-mode, and I dare not guess which way the comparison would favour. I for one haven't missed ex-mode yet.

[...]

PR #624, opened 2015-04-13
Repeating search with //, ?? et al.

### LINKED ISSUE #578: // matches literal /

jacekkopecky:
fixes #578 
Now search can be repeated with `/<enter>`, `//<enter>`, `?<enter>` and `??<enter>`.
Beyond that, I think I'm beginning to agree that / should be handled by ex-mode (#596).


maxbrunsfeld:
Awesome.

PR #634, opened 2015-04-24
add command shortcuts for save, close, and find/replace

lehresman:
I'm loving vim-mode, but my fingers keep wanting to do ":w" and ":q" to save and close a pane.  This adds those key mappings.  I also added a "/" key mapping that brings up the find/replace panel.


daniloisr:
Hi @lehresman, you can install https://atom.io/packages/ex-mode to have the `":"` command mode. :smile:

lehresman:
Wow, perfect!  That's much better than my solution.  Thanks for the tip!

This PR can probably be closed since it's irrelevant.

Issue #517, opened 2015-02-11
Add text objects for paragraph ('ip' / 'ap')


nesmyslny:
It's really awesome to have a vim mode in atom. So thanks to all contributing to this plugin!

I'm missing some vim features, that I use quite often. One of them is the paragraph text object:
* `ap`: a paragraph
* `ip`: inner paragraph

For example, `vap` selects the whole current paragraph. Here's the documentation: http://vimhelp.appspot.com/motion.txt.html#ap

Are there plans to support this? Maybe I find some time, to look at it (though I'm totally new to Atom).

**Update 2015-05-13**
* `ip` was implemented by @t9md (https://github.com/t9md/vim-mode/commit/4592680a423ea134b743ca3be027f86c9e4b80b8)
* `ap` behaves like `ip` (https://github.com/atom/vim-mode/issues/517#issuecomment-101594821). So to be precicely, `ap` is not implemented yet.

nesmyslny:
:sparkles: Thanks @t9md! :bow: 
Is this issue still open by purpose or should I close it?

t9md:
Currently `ip` and `ap` both work as `ip`. 
So to be precicely, `ap` is still not implemented yet. 
I rarely use ap in Vim.
I don't have motivation to implement `ap` for now, since I'm very new to Atom itself(about 2 week), and need very basic understanding of Atom itself and vim-mode architecture.

nesmyslny:
Ok, I see. I didn't noticed, because me neither use `ap` :wink:
Maybe someone has interest in it and finds this issue (so I don't close it).

Thanks again!

MarkusSN:
I think this can closed :)

maxbrunsfeld:
Yup, thanks @MarkusSN!

PR #639, opened 2015-04-28
support textobj paragraph #517

t9md:
I added textobj paragraph feature, currently.

 - no difference for `ap` and `ip`.
 - work on multiple cursor
 - behavior on blank line is different from `Vim`. Vim select consecutive blank line, but this won't select any range.


maxbrunsfeld:
This looks good. Could you add a test for this? Let me know if you have any questions about our tests.

t9md:
I only add spec for `ip`(inside-paragraph).
I'm not sure I should add spec for `ap`(around-paragraph), since currently identical behavior to `ip`.

maxbrunsfeld:
> I'm not sure I should add spec for ap(around-paragraph), since currently identical behavior to ip.

I agree. The spec you added looks good. Thanks for adding this!

PR #643, opened 2015-04-29
add configuration setting for ctrl-a/x number regexps

### LINKED ISSUE #641: Better increase/decrease handling for negative number.
Recently increase/decrease feature is supported by PRs on #636.
That was great! thanks!!
When I use increase number by `ctrl-a`. I frequently hit following situation.

Editing `keymap.cson`, I copied last line from above line and increase number to `z 2`,
```cson
'atom-text-editor.vim-mode.command-mode':
  'z 1':    'editor:fold-at-indent-level-1'
  'z 2':    'editor:fold-at-indent-level-1'
```
Then move to right by `l` and increase next number, it result in 
```cson
  'z 2':    'editor:fold-at-indent-level0'
```
I want `editor:fold-at-indent-level-2` in this case.
In this case I don't want it treated as negative number when `-` is preceded by like [A-Za-z].
This is not feature original Vim supported, but I think this is reasonable enhancement.
So when increase/decrease pick up number, wether it take minus char(`-`) as manipulation target should be determined by the preceding character.

- Expected behavior by this enhancement.
```
# treat this as positive number
"foo-bar-1" # c-a => foo-bar-2 , c-x foo-bar-0
```

Since it break expectation for original Vim user, it might better to provide this feature as optional configuration.

jacekkopecky:
fixes #641

### DIFF
--- a/lib/operators/increase-operators.coffee
+++ b/lib/operators/increase-operators.coffee
@@ -1,16 +1,17 @@
 {Operator} = require './general-operators'
 {Range} = require 'atom'
+settings = require '../settings'
 
 #
 # It increases or decreases the next number on the line
 #
 class Increase extends Operator
   step: 1
-  numberRegex: new RegExp("-?[0-9]+")
 
   constructor: ->
     super
     @complete = true
+    @numberRegex = new RegExp(settings.numberRegex())
 
   execute: (count=1) ->
     @editor.transact =>
--- a/lib/settings.coffee
+++ b/lib/settings.coffee
@@ -13,6 +13,10 @@ settings =
     useClipboardAsDefaultRegister:
       type: 'boolean'
       default: false
+    numberRegex:
+      type: 'string'
+      default: '-?[0-9]+'
+      description: 'Use this to control how Ctrl-A/Ctrl-X finds numbers; use "(?:\\B-)?[0-9]+" to treat numbers as positive if the minus is preceded by a character, e.g. in "identifier-1".'
 
 Object.keys(settings.config).forEach (k) ->
   settings[k] = ->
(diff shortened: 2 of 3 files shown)


lib/operators/increase-operators.coffee
```diff
@@ -1,16 +1,17 @@
...
 # It increases or decreases the next number on the line
 #
 class Increase extends Operator
   step: 1
-  numberRegex: new RegExp("-?[0-9]+")
 
   constructor: ->
     super
     @complete = true
+    @numberRegex = new RegExp(settings.numberRegex() or settings.config.numberRegex.default)
```
maxbrunsfeld: I think you don't need to explicitly fall back on the default value; [`atom.config.get`](https://atom.io/docs/api/v0.196.0/Config#instance-get) should return the default value if the user has not overridden it.

jacekkopecky: For some reason the tests won't run for me without that.

t9md:
Thanks @jacekkopecky , for me, I tested your code by copy&paste, as @maxbrunsfeld said, explicit fallback wasn't necessary.
It might test(spec) related issue(since I didn't run spec) or your environment issue.

Anyway thanks for this, and I hope this PR will be merged.

jacekkopecky:
@t9md without the fallback, the new specs were failing for me – it seems when running specs, `atom.config.get` doesn't return the proper defaults. I don't have the brain cycles now to look into that, so I've included a workaround in the specs now instead of in the main code.
@maxbrunsfeld – is this better now?

maxbrunsfeld:
:+1: Thanks!

PR #653, opened 2015-05-07
Add short highlight when text is yanked

DaanDD:
When text gets yanked, a short highlight appears as a small visual aid. Inspired by the Vintageous package from Sublime Text.

The only problem is that I could not figure out how to write specs for this feature. The specs in this [example repo](https://github.com/atom/decoration-example/blob/master/spec/decoration-example-spec.coffee) all seem to be using the Atom WorkspaceView, but the operators-spec.coffee seems to be using another way to initialise the editor. I also got quite some deprecation warnings while running the package specs.

So this feature should work, but there are no specs yet.

### DIFF
--- a/lib/operators/general-operators.coffee
+++ b/lib/operators/general-operators.coffee
@@ -200,6 +200,11 @@ class Yank extends Operator
 
     @setTextRegister(@register, text)
 
+    marker = @editor.markBufferRange(@editor.getSelectedBufferRange(), invalidate: 'touch')
+    yankHighlight = @editor.decorateMarker(marker, type: 'highlight', class: 'yank-highlight')
+
+    setTimeout (-> yankHighlight.destroy()), 300
+
     @editor.setSelectedBufferRanges(newPositions.map (p) -> new Range(p, p))
     @vimState.activateCommandMode()
 
--- a/styles/vim-mode.less
+++ b/styles/vim-mode.less
@@ -56,3 +56,11 @@ atom-text-editor.vim-mode.visual-mode
     }
   }
 }
+
+atom-text-editor,
+atom-text-editor::shadow {
+  .yank-highlight .region {
+    border-radius: @component-border-radius;
+    border: 1px solid @syntax-result-marker-color;
+  }
+}


bronson:
I love this.

Personally I'd like the highlight to be briefer.  Might be worth making it a configurable setting?  That would also offer and obvious way to turn it off.

`2yy` draws a box around the following line, making it look like only one line got yanked.  Hitting `p` shows that the yank worked correctly and it's the highlight that was misleading.

Also `yiw`, `yy`, etc only show underline highlights, not boxes.

I notice that highlighting multiple-line selections can look weird:

bug3

If there's a way to make all of these boxes, so it's clear exactly what got yanked, I think this feature would be phenomenal.

bronson:
Also, I have a similar testing issue in https://github.com/atom/vim-mode/pull/723.  (Well, I'm switching tabs so our tests are different, but being unable to figure out how to write UI tests is the same.)

Hoping @maxbrunsfeld has some time to offer a hint or two on the right way to do it.

bronson:
@t9md did you mention on slack that you were having similar issues with quick-highlight?  Any chance this is an an easy fix?

t9md:
@bronson This is not same issue I talked on slack.
I think its because of decoration is line-wise?
Simply use style othe than box I mean style without border.

t9md:
My package clip-history do similar decoration to highlight pasted area. See gif anime of clip-history to chek how its like.

t9md:
I myself dont want to include these type of fancy feature to vim-mode core.
Maybe it could be disabled by configuration but it increase lines of code and config options.
Let vim-mode provide consistent event hook like onDidYank(callback) and callback passed range and text just yanked.
And user use these hook in other vim-mode-decorativelanguage.

DaanDD:
@t9md: That seems like a good idea. I always found it really handy in Sublime Text's vim mode.

bronson:
@t9md I had the same thought as you when I first looked at this patch.  I only merged it just to be complete.  \*I\* sure didn't want it.

... but I've grown to really like it.  Maybe I'm losing my edge.

For example, I immediately see that I should have typed `yaw` instead of `yiw`.  The days of navigating to the new buffer, pasting, seeing I got the yank wrong, navigating back to the original buffer, and trying again are gone!  Immediate feedback is really nice.

In general, I agree that vim-mode should only include core vim stuff.  But this is 3 lines of code...  As long as it's turned off by default, I like the idea of including it.

A hook would be better of course, but not because it saves any code.

bronson:
I removed this patch from vim-mode-next...  A hook is the right way to do this.

t9md:
I like this kind of immediate feedback decoration.
But if the more we add vim-mode this kind of decoration, the more core code complicated.
And these preference is greatly vary from person to person, it sometimes become difficult to reach conclusion.
So hook is better for
 - keep core code simple
 - extensible for other package.

So I think we should open new Issue to discuss what type of hook(or event) is required.
For this topic, `onDidYank(callback)` and callback's argument should `calback({text, range})` 
and its event is fired on `y` family operation.

bronson:
And, to completely hijack this PR, I hope whoever investigates adding highlight hooks keeps #432 (highlighting find-previous and find-next jumps) in mind too.

maxbrunsfeld:
I like this feature too, and I'd be ok with adding it similarly to how it's currently implemented (as opposed to adding some kind of hook) but just *without* the css change. That way, users and theme authors could customize the look of the highlight just by tweaking CSS. We should also make the hard-coded timeout short so that the duration can be customized via css transitions.

There is some code in Atom core for doing things like this, but it is not yet documented. It's called 'flashing' a decoration. See [this code](https://github.com/atom/find-and-replace/blob/7c678111b948b442ba2fa7cdbd7882eb5b9422ab/lib/find-view.coffee#L368) in `find-and-replace` and [these private methods](https://github.com/atom/atom/blob/2e2fef0c476d55bf283abafb10bf53b8f6386e75/src/decoration.coffee#L171) on the `Decoration` class.

If anyone is interested in taking this further, I would suggest trying this:

```coffee
  selection.setBufferRange(selection.getBufferRange(), flash: true)
```

You could also try flashing a custom decoration:

```coffee
marker = @editor.markBufferRange(range)
decoration = @editor.decorateMarker(marker, type: 'highlight', class: 'yank-highlight')
decoration.flash('yank-flash', 50)
```

Then maybe tweaking the CSS to get the flash looking nice.

lee-dohm:

PR #654, opened 2015-05-07
fix deprecation of editor.getUri

jacekkopecky:
with `apm test -1` I've uncovered another use of a deprecated method, this fixes it.
How does one add `-1` to `apm test` in Travis tests?

### DIFF
--- a/lib/vim-state.coffee
+++ b/lib/vim-state.coffee
@@ -276,7 +276,7 @@ class VimState
       type = Utils.copyType(text)
       {text, type}
     else if name is '%'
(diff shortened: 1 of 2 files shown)


kevinsawicki:
> How does one add -1 to apm test in Travis tests?

It was added today for all packages using that standard `travis.yml` file from https://github.com/atom/ci/blob/master/.travis.yml via https://github.com/atom/ci/pull/13

kevinsawicki:
Thanks for fixing these!

jacekkopecky:
Perfect!

Issue #645, opened 2015-04-30
Blockwise-visual-mode not working


TheLuckyOne:
Hi! 
Vim-mode:activate-blockwise-visual-mode not working on my mac, linux and windows. 
What am i doing wrong?

bronson:
Get this, Atom doesn't support block select yet: https://github.com/atom/atom/issues/2306

Apparently there's been some work but nothing usable yet.  If Atom 1.0 ships without block select, that would be downright embarrassing.

bronson:
Also, this is a dupe of https://github.com/atom/vim-mode/issues/516

Issue #468, opened 2014-12-09
Implement visual block mode


gosukiwi:
I can't seem to find an answer to this so sorry if it's duplicated, but I'd like to know if visual block mode is implemented, I tried setting a `ctrl-q` key binding to `blockwise-visual-mode` but I just get regular visual mode activated.

``` cson
'.editor.vim-mode.command-mode':
  'ctrl-v': 'vim-mode:activate-blockwise-visual-mode'
```

maxbrunsfeld:
No, visual block mode is not currently implemented. It should be coming soon.

cpadilla:
How soon is soon? Any updates?

sudostack:
+1

jeffschwartz:
+1

ahmadseleem:
+1

irrationalistic:
Hey, thought I'd link to this other package that seems to be pretty close to this feature. Not sure if there's any collaboration that can happen! https://github.com/k-yamada/block-selection-mode

fbob:
To me this is the killer feature which prevent me to switch full time to atom
Anyway, great job guys :)
Thanks !

paulcsmith:
I also would love to see this :+1:

femueller:
+1

rtsolv:
:+1:

pthrasher:
I've got a basic (read: incomplete) implementation started here: #658 

I'd love to get some feedback / possible collaboration.

t9md:
Here is dirty hack to emulate visual-block-mode using multi-cursor.
Not perfect solution, but this cover basic visual-block use case for me.

https://gist.github.com/t9md/736291fd22260b5c7de0

ahmadseleem:
This is the best I found so far:
https://atom.io/packages/column-select

Still missing a toggle like your solution...
hopefully this could be ported to vim-mode here...

t9md:
After I put above gist, I just started developing package and just released now.

https://atom.io/packages/vim-mode-visual-blockwise

This is dirty hack of about 4 hour, there must be a lot of improvement but now I'm sleepy...

Anyway, I haven't checked or used `column-select`, will check afterward.

rodmax:
"""
To me this is the killer feature which prevent me to switch full time to atom
Anyway, great job guys :)
"""
 nothing more to add
+1

shemerey:
+1 for this feature

jacekkopecky:
[...]

PR #658, opened 2015-05-08
[WIP] add basic/incomplete blockwise selection handling

pthrasher:
This is a work-in-progress. I'm hoping to get some feedback / collaboration.

This is just my first pass at getting blockwise selection to work.

This implementation fails when down-arrowing into an empty line, and also, currently there's not a simple way to track the cursor when moving up and down since atom doesn't let you put the cursor places where there's no text (afaik).
#645
#468

atom/atom#2306

### DIFF
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -17,6 +17,10 @@ class Motion
   constructor: (@editor, @vimState) ->
 
   select: (count, options) ->
+    if @isBlockwise()
+      @moveSelectionBlockwise(count, options)
+      return false
+
     value = for selection in @editor.getSelections()
       if @isLinewise()
         @moveSelectionLinewise(selection, count, options)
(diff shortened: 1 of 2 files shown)


bronson:
Just curious, doesn't it make sense for Atom core to implement blockwise selection first?  It looks like your technique will handle the text pretty well all on its own, but the visual appearance will be pretty difficult to get right.  (pure speculation, tell me if I'm wrong).

That said, since blockwise select isn't on Atom's 1.0 punchlist (!!), maybe it does make sense just to do it here.  It's an absolute requriement for any serious text editor.

pthrasher:
@bronson Yes, it does. First, atom/text-buffer will need to support out-of-bounds selections. (beyond line endings)

After that, it should be easy to add the visual changes to atom proper. I opened an issue to begin discussing this a while back, but haven't heard anything. (atom/text-buffer#59)

I've been swamped, and haven't come back to this, but I will hopefully clean up this version and fix some kinks this week. I think having *something* in place is a good first start. After that, perhaps I can take a whack at adding out-of-bounds support to text-buffer and see what people think.

maxbrunsfeld:
> currently there's not a simple way to track the cursor when moving up and down.

@pthrasher in atom, cursors have the notion of a 'goal column', similar to other text editors, such that when you move the cursor vertically from a long line onto a shorter one and back again, it will return to the original column. You shouldn't have to do anything special to make this work; the normal cursor movement commands should take care of it.

bronson:
There's this too: https://atom.io/packages/vim-mode-visual-block

No yank or paste yet.

lee-dohm:

PR #677, opened 2015-05-24
make dd leave cursor on nonblank char

### LINKED ISSUE #315: 'dd' should leave cursor at first non whitespace character

bronson:
Default Vim behavior is to move to the first nonblank character in the line.
Also, here seems to be a plugin that causes Vim to leave the cursor in the
same column if possible.  vim-mode's old behavior did neither.

This PR makes vim-mode work just like stock vim.

### DIFF
--- a/spec/motions-spec.coffee
+++ b/spec/motions-spec.coffee
@@ -677,8 +677,7 @@ describe "Motions", ->
 
         it "deletes the current line plus that many previous lines", ->
           expect(editor.getText()).toBe "1\n6\n"
-          # commented out because the column is wrong due to a bug in `k`; re-enable when `k` is fixed
-          #expect(editor.getCursorScreenPosition()).toEqual [1, 0]
+          expect(editor.getCursorScreenPosition()).toEqual [1, 0]
 
   describe "the + keybinding", ->
     beforeEach ->
@@ -738,8 +737,7 @@ describe "Motions", ->
 
         it "selects to the first character of the next line", ->
           expect(editor.getText()).toBe "abcdefg\n"
-          # commented out because the column is wrong due to a bug in `j`; re-enable when `j` is fixed
-          #expect(editor.getCursorScreenPosition()).toEqual [0, 0]
+          expect(editor.getCursorScreenPosition()).toEqual [0, 0]
 
     describe "with a count", ->
       beforeEach ->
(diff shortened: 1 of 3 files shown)


spec/motions-spec.coffee
```diff
@@ -633,9 +632,10 @@ describe "Motions", ->
           keydown('-')
 
         it "selects to the first character of the previous line (directly above)", ->
-          expect(editor.getText()).toBe "abcdefg\n"
+          # commented out because vim does not put a newline but vim-mode does
+          #expect(editor.getText()).toBe "abcdefg"
```
maxbrunsfeld: Did your change cause this test to fail? I can't tell why this would be affected by skipping the leading whitespace.

spec/operators-spec.coffee
```diff
@@ -328,6 +328,17 @@ describe "Operators", ->
...
+        editor.setText("12345\n  abcde\n")
+        editor.setCursorScreenPosition([0, 4])
+
+        keydown('d')
+        keydown('d')
+
+        expect(editor.getText()).toBe "  abcde\n"
+        expect(editor.getCursorScreenPosition()).toEqual [0, 2]
+
+
```
maxbrunsfeld: Could you remove this second blank line? Very minor, but seems worth mentioning since it looks like you're doing a lot of work on this package :+1:.

maxbrunsfeld:
Looks good except for the comments above.

spec/motions-spec.coffee
bronson: I pulled these commits out of a hairier patchset fixing https://github.com/atom/vim-mode/pull/386 and accidentally let this line come along for the ride.

Technically, this assertion is wrong -- vim doesn't put the newline.  But that's a fix for another day.

spec/motions-spec.coffee
```diff
@@ -652,7 +651,8 @@ describe "Motions", ->
           keydown('-')
 
         it "selects to the first character of the previous line", ->
-          expect(editor.getText()).toBe "abcdefg\n"
+          # commented out because vim does not put a newline but vim-mode does
+          #expect(editor.getText()).toBe "abcdefg"
```
maxbrunsfeld: Could you restore this too?

spec/motions-spec.coffee
```diff
@@ -699,7 +698,8 @@ describe "Motions", ->
           keydown('+')
 
         it "deletes the current and next line", ->
-          expect(editor.getText()).toBe "  abc\n"
+          # commented out because vim does not put a newline but vim-mode does
+          #expect(editor.getText()).toBe "  abc"
```
maxbrunsfeld: and this

spec/motions-spec.coffee
bronson: sure.  what's worse: missing specs, or incorrect specs?  :)

maxbrunsfeld: This is fine, but for future PRs, it'd be better to not add comments on things unrelated to the PR. It would be better to leave those lines unchanged and then possibly submit a separate PR that fixes the other bug.

bronson:
@maxbrunsfeld I agree with everything you said so I rebased that other stuff away.  That was getting a little silly.

Anything more I can do?

maxbrunsfeld:
Thanks!

PR #684, opened 2015-05-28
Add the ap text-object, modify the ip text-object,

### LINKED ISSUE #674: `dap` doesn't delete newlines
Before `dap` is typed:

    paragraph 1, line 1█
    p1, line 2

    paragraph 2, line 1
    p2, line 2

Expected result after `dap` is typed:

    █paragraph 2, line 1
    p2, line 2

Actual result after `dap` is typed:

    █

    paragraph 2, line 1
    p2, line 2

This bug really makes it painful to delete multiple paragraphs in a row, or to paste the deleted paragraphs somewhere else (you end up having to add and delete a bunch of empty lines).

MarkusSN:
This should fix #674. I also ended up changing the inner paragraph text-object after I noticed that it didn't work as in vim. Now it seems to behave as in vim, which I am assuming is desirable.

### DIFF
--- a/lib/text-objects.coffee
+++ b/lib/text-objects.coffee
@@ -147,6 +147,18 @@ class SelectInsideParagraph extends TextObject
       range = selection.cursor.getCurrentParagraphBufferRange()
       if range?
         selection.setBufferRange(range)
+        selection.selectToBeginningOfNextParagraph()
       true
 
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
Great work; thanks so much!

Issue #685, opened 2015-05-30
open vertical and horizontal splits from tree view using vim shortcuts


sheldonkwok:

bronson:
You mean typing <kbd>V</kbd> when the tree view is focused would open the selected file in a vertical split?  Same for H and horizontal split?

sheldonkwok:
ctrl-w v for vertical and ctrl-w s for horizontal

bronson:
Ah, I thought you were asking for nerdtreeish functionality.  You'd just like Vim's regular bindings to apply there.  Makes sense.

That invites the question of "how much monkeypatching of other packages is acceptable for vim-mode?"

bronson:
One way to solve it might be to create a `tree-view-vim-mode` package that layers vim's keybindings on top of tree-view.  That at least isolates the monkeypatching into a small, easy-to-disable package.  Hopefully it wouldn't explode into an n+1 problem where there's a _package_-vim-mode for every installed package.  :)

sheldonkwok:
Ah that works too. Should've noticed that the modifications were for the editor only. Thanks!

lee-dohm:
As stated in [the README](https://github.com/atom/vim-mode), this package is no longer maintained and is deprecated. We recommend that people use the [vim-mode-plus package](https://atom.io/packages/vim-mode-plus) instead. Because of this, we are [archiving this repository](https://help.github.com/articles/about-archiving-repositories/) and closing all issues and pull requests. Thanks very much for your support and contributions!

PR #686, opened 2015-05-30
add tree view vertical/horizontal split bindings

sheldonkwok:
https://github.com/atom/vim-mode/issues/685

### DIFF
--- a/keymaps/vim-mode.cson
+++ b/keymaps/vim-mode.cson
@@ -321,3 +321,7 @@
   '" +': 'vim-mode:register-prefix'
   '" %': 'vim-mode:register-prefix'
   '" _': 'vim-mode:register-prefix'
+
+'.tree-view':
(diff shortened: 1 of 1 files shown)


bronson:
I'm a little worried about the precedent this sets...  Should tree-view support ctrl-w ctrl-h?  How about / to search?

That said, I think this is great.  There doesn't seem to be any real downside.  :+1:

bronson:
After further thought, I've changed my mind...  This is the perfect use for `~/.atom/keymap.cson`.

I do think it does make sense to collect these ideas into a docs file and include them with vim-mode.  For now, I added an Interesting Ideas section to [v-m-n's readme](https://github.com/bronson/vim-mode-next#interesting-ideas).

sheldonkwok:
Sounds good, thanks for the input

PR #692, opened 2015-06-05
don't moveLeft() if selection.isReversed()

t9md:
Before this change, `deactivateVisualMode()` unnecessarily move cursor Left when selection is Reversed.

## Reproduce
* `v` to enter visual-mode
* `kk`, `esc` move Cursor one column **Left** position to expected position.

This PR fix above bug.
All test pass, I don't think additional test required for this fix.


maxbrunsfeld:
Thanks for catching that!

PR #696, opened 2015-06-06
Fix for #689 and improve blank line handling for the ip/ap text-objects

### LINKED ISSUE #689: vim-mode delimits paragraphs differently to Vim
Before `dap` is typed:

    # XXX fix
    add_atom_command: (binding, fn) ->
        binding = 'gpd:' + binding
        █@subscriptions.add atom.commands.add 'atom-workspace', binding: -> fn.call(@)

Expected result after `dap` is typed:

    █

Actual result after `dap` is typed:

    # XXX fix
    █

Not sure how important this is, just thought I'd report it anyway.

MarkusSN:
Fixes #689. Also makes the handling of blank lines, and lines that only contain whitespace more like in vim, e.g. you can now select the lines between two paragraphs by doing 'vip' .

### DIFF
--- a/spec/text-objects-spec.coffee
+++ b/spec/text-objects-spec.coffee
@@ -193,7 +193,6 @@ describe "TextObjects", ->
       editor.setCursorBufferPosition([2, 2])
 
     it "applies operators inside the current paragraph in operator-pending mode", ->
-
       keydown('y')
       keydown('i')
       keydown('p')
@@ -205,7 +204,6 @@ describe "TextObjects", ->
       expect(editorElement.classList.contains('normal-mode')).toBe(true)
 
     it "selects inside the current paragraph in visual mode", ->
-
       keydown('v')
       keydown('i')
       keydown('p')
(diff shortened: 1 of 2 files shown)


MarkusSN:
Any feedback on this?

bronson:
I think it looks great.  Definitely works better now.

I notice that hitting undo after 'dap' or 'dip' leaves the cursor at the end of the deleted region.  Vim leaves the cursor at the beginning of the deleted region.  Might be worth opening a new issue for this though, not sure how deep this problem runs.

Ideally you'd combine some of these commits so the changes are more obvious.

jacekkopecky:
Seems to need rebasing please.

MarkusSN:
Refactored and rebased.

jacekkopecky:
please remove `.swp` from the commit :sunglasses:

MarkusSN:
Sneaky .swp removed :)

spec/text-objects-spec.coffee
```diff
@@ -205,15 +205,36 @@ describe "TextObjects", ->
       expect(editorElement.classList.contains('normal-mode')).toBe(true)
 
     it "selects inside the current paragraph in visual mode", ->
+
```
jacekkopecky: this seems to be an unnecessary empty line; with or without it I'm still recommending an immediate merge for this PR 8-)

maxbrunsfeld:
Great work @MarkusSN; thanks for fixing this!

PR #710, opened 2015-06-24
Multiple cursor undo of delete operations

jacekkopecky:
Delete operations with multiple cursors are undone one by one. They should be undone all together.


bronson:
absolutely right, works great for me.

PR #711, opened 2015-06-24
Fix p cursor position

jacekkopecky:
If you have a word in the clipboard and `p` it at the end of the line, the cursor ends up on the second-to-last character of the word, not on the last one, as expected. This PR fixes that.


bronson:
agreed.  bug is 100% reproducible, fix looks great.

jacekkopecky:
rebased, low-hanging fruit

Issue #650, opened 2015-05-04
Repeat doesn't interact correctly with autocomplete


podgib:
When repeating a previous command using '.', the repeated command doesn't include characters added by autocomplete.

Eg.
```C
  int x = f(input);
  int y = f(input);
```

If I move my cursor to the start of 'input', then press cw, 'input' is removed and I am in insert mode. I then start typing 'output', but after 'out', I use autocomplete to finish the word. If I now leave insert mode, move to the next 'input' and press '.', it only inserts 'out', not 'output'.

Expected:
```C
  int x = f(output);
  int y = f(output);
```

Actual:
```C
  int x = f(output);
  int y = f(out);
```

jacekkopecky:
Yes, and it also doesn't play nice with automatic matching of parentheses:
`i(abc)<esc>` can replay as "()abc". I haven't yet had enough time to try
and solve this one. I expect one fix can cover both problems, though.

maxbrunsfeld:
I think this was fixed by #712.

Issue #564, opened 2015-03-11
insert repeat gets confused by `delete`


jacekkopecky:
in insert mode, `delete` will remove the character following the cursor, but `.` doesn't replay it correctly. Demonstration of the problem: 
 1. have `1234` in the editor and the cursor is on 2
 2. type `iabc<delete><esc>`  – you should have `1abc34` in the editor and the cursor on `c`
 3. move cursor on the `1`
 4. type `.` – you should have `abcabc34` in the text editor, with cursor on the first `c` (the original `1` would get eaten by the repeated `delete`)

But what you actually get is `ab1abc34` with cursor on the first `b`.

The problem is in `TransactionBundler` in `input.coffee` which treats the `delete` as a `backspace`.

maxbrunsfeld:
Thanks for fixing this.

Issue #392, opened 2014-08-27
Repeat (.) while matching completion is enabled can lead to erroneous inserts


greg-hellings:
I was replacing a word of text with a method invocation in a PHP file. To do this, I was in command mode and press "c-w", then enter the text which involved both a method invocation and a string literal, for example:

method('mystring')

I then moved to another place where I wished to make the same substitution and pressed ".". The resulting text was

method()''mystring

PHP is configured to auto-close parentheses and quotation marks, likely leading to this strange behavior.

peterjeliot:
I'm experiencing the same bug with quotation marks in ruby.

maxbrunsfeld:
I believe this was fixed by #712.

PR #712, opened 2015-06-25
Fix insert redo

jacekkopecky:
This fixes the problem where bracket-matcher's autocompletion of quotes or brackets would confuse replay of inserts. This is mentioned in #392, #650, and related to #564.

### DIFF
--- a/lib/operators/input.coffee
+++ b/lib/operators/input.coffee
@@ -12,7 +12,7 @@ class Insert extends Operator
   isComplete: -> @standalone or super
 
   confirmChanges: (changes) ->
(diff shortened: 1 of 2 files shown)


jacekkopecky:
It seems that both #392 and #650 are actually fixed by this PR (but then, I am tired, so @greg-hellings, @peterjeliot, @podgib please verify).

jacekkopecky:
Right, now it's also handling autocomplete.

maxbrunsfeld:
Wonderful. Thanks so much.

peterjeliot:
Very nice! I couldn't break it.

Issue #713, opened 2015-06-25
Uncaught NotFoundError: Failed to execute 'remove' on 'Element': The node to be removed is no lon...


jbiason:
[Enter steps to reproduce below:]
1. ...
2. ...

**Atom Version**: 0.211.0
**System**: Ubuntu 14.04.2
**Thrown From**: [vim-mode](https://github.com/atom/vim-mode) package, v0.53.0
### Stack Trace

Uncaught NotFoundError: Failed to execute 'remove' on 'Element': The node to be removed is no longer a child of this node. Perhaps it was moved in a 'blur' event handler?

```
At /home/julio.biason/.atom/packages/vim-mode/lib/view-models/view-model.coffee:8

Error: Failed to execute 'remove' on 'Element': The node to be removed is no longer a child of this node. Perhaps it was moved in a 'blur' event handler?
  at Error (native)
  at /home/julio.biason/.atom/packages/vim-mode/lib/view-models/view-model.coffee:8:43
  at Emitter.module.exports.Emitter.emit (/home/julio.biason/.atom/packages/vim-mode/node_modules/event-kit/lib/emitter.js:71:11)
  at VimState.module.exports.VimState.pushOperations (/home/julio.biason/.atom/packages/vim-mode/lib/vim-state.coffee:213:18)
  at atom-text-editor.commands.(anonymous function) (/home/julio.biason/.atom/packages/vim-mode/lib/vim-state.coffee:196:45)
  at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/opt/atom/resources/app.asar/src/command-registry.js:241:29)
  at /opt/atom/resources/app.asar/src/command-registry.js:3:61
  at KeymapManager.module.exports.KeymapManager.dispatchCommandEvent (/opt/atom/resources/app.asar/node_modules/atom-keymap/lib/keymap-manager.js:520:16)
  at KeymapManager.module.exports.KeymapManager.handleKeyboardEvent (/opt/atom/resources/app.asar/node_modules/atom-keymap/lib/keymap-manager.js:355:22)
  at HTMLDocument.module.exports.WindowEventHandler.onKeydown (/opt/atom/resources/app.asar/src/window-event-handler.js:180:20)

```
### Commands

```
     -0:24.9.0 vim-mode:activate-command-mode (atom-text-editor.editor.vim-mode.is-focused.insert-mode)
[...]

Issue #646, opened 2015-04-30
Uncaught NotFoundError: Failed to execute 'remove' on 'Element': The node to be removed is no longer a child of this node. Perhaps it was moved in a 'blur' event handler?


patrickmedaugh:
[Enter steps to reproduce below:]
1. ...
2. ...

**Atom Version**: 0.189.0
**System**: Mac OS X 10.10.2
**Thrown From**: [vim-mode](https://github.com/atom/vim-mode) package, v0.43.0
### Stack Trace

Uncaught NotFoundError: Failed to execute 'remove' on 'Element': The node to be removed is no longer a child of this node. Perhaps it was moved in a 'blur' event handler?

```
At /Users/pmedaugh/.atom/packages/vim-mode/lib/view-models/view-model.coffee:8

Error: Failed to execute 'remove' on 'Element': The node to be removed is no longer a child of this node. Perhaps it was moved in a 'blur' event handler?
  at Error (native)
  at /Users/pmedaugh/.atom/packages/vim-mode/lib/view-models/view-model.coffee:8:43
  at Emitter.module.exports.Emitter.emit (/Users/pmedaugh/.atom/packages/vim-mode/node_modules/event-kit/lib/emitter.js:71:11)
  at VimState.module.exports.VimState.pushOperations (/Users/pmedaugh/.atom/packages/vim-mode/lib/vim-state.coffee:209:18)
  at atom-text-editor.commands.(anonymous function) (/Users/pmedaugh/.atom/packages/vim-mode/lib/vim-state.coffee:192:45)
  at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/Applications/Atom.app/Contents/Resources/app/src/command-registry.js:242:29)
  at /Applications/Atom.app/Contents/Resources/app/src/command-registry.js:3:61
  at KeymapManager.module.exports.KeymapManager.dispatchCommandEvent (/Applications/Atom.app/Contents/Resources/app/node_modules/atom-keymap/lib/keymap-manager.js:552:16)
  at KeymapManager.module.exports.KeymapManager.handleKeyboardEvent (/Applications/Atom.app/Contents/Resources/app/node_modules/atom-keymap/lib/keymap-manager.js:392:22)
  at HTMLDocument.module.exports.WindowEventHandler.onKeydown (/Applications/Atom.app/Contents/Resources/app/src/window-event-handler.js:180:20)

```
### Commands

```
[...]

Issue #642, opened 2015-04-29
Error: Failed to execute 'remove' on 'Element':


LukeMauldin:
**Atom Version**: 0.194.0
**System**: Mac OS X 10.10.3
**Thrown From**: [vim-mode](https://github.com/atom/vim-mode) package, v0.45.0
### Stack Trace

Uncaught NotFoundError: Failed to execute 'remove' on 'Element': The node to be removed is no longer a child of this node. Perhaps it was moved in a 'blur' event handler?

```
At /Users/lmauldin/.atom/packages/vim-mode/lib/view-models/view-model.coffee:8

Error: Failed to execute 'remove' on 'Element': The node to be removed is no longer a child of this node. Perhaps it was moved in a 'blur' event handler?
  at Error (native)
  at /Users/lmauldin/.atom/packages/vim-mode/lib/view-models/view-model.coffee:8:43
  at Emitter.module.exports.Emitter.emit (/Users/lmauldin/.atom/packages/vim-mode/node_modules/event-kit/lib/emitter.js:71:11)
  at VimState.module.exports.VimState.pushOperations (/Users/lmauldin/.atom/packages/vim-mode/lib/vim-state.coffee:207:18)
  at atom-text-editor.commands.(anonymous function) (/Users/lmauldin/.atom/packages/vim-mode/lib/vim-state.coffee:190:45)
  at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/Applications/Atom.app/Contents/Resources/app.asar/src/command-registry.js:238:29)
  at /Applications/Atom.app/Contents/Resources/app.asar/src/command-registry.js:3:61
  at KeymapManager.module.exports.KeymapManager.dispatchCommandEvent (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/atom-keymap/lib/keymap-manager.js:519:16)
  at KeymapManager.module.exports.KeymapManager.handleKeyboardEvent (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/atom-keymap/lib/keymap-manager.js:354:22)
  at HTMLDocument.module.exports.WindowEventHandler.onKeydown (/Applications/Atom.app/Contents/Resources/app.asar/src/window-event-handler.js:178:20)

```
### Commands

```
  2x -0:07.4.0 vim-mode:delete (atom-text-editor.editor.vim-mode.is-focused.command-mode)
[...]

Issue #633, opened 2015-04-24
Error on normal mode


gokmen:
In `normal` ( or `command` for this package) mode when I press `c` `m` it crashes with following stack trace.

**Atom Version**: 0.194.0
**System**: Mac OS X 10.10.3
**Thrown From**: [vim-mode](https://github.com/atom/vim-mode) package, v0.44.0

### Stack Trace

Uncaught NotFoundError: Failed to execute 'remove' on 'Element': The node to be removed is no longer a child of this node. Perhaps it was moved in a 'blur' event handler?

```
At /Users/gokmen/.atom/packages/vim-mode/lib/view-models/view-model.coffee:8

Error: Failed to execute 'remove' on 'Element': The node to be removed is no longer a child of this node. Perhaps it was moved in a 'blur' event handler?
  at Error (native)
  at /Users/gokmen/.atom/packages/vim-mode/lib/view-models/view-model.coffee:8:43
  at Emitter.module.exports.Emitter.emit (/Users/gokmen/.atom/packages/vim-mode/node_modules/event-kit/lib/emitter.js:71:11)
  at VimState.module.exports.VimState.pushOperations (/Users/gokmen/.atom/packages/vim-mode/lib/vim-state.coffee:209:18)
  at atom-text-editor.commands.(anonymous function) (/Users/gokmen/.atom/packages/vim-mode/lib/vim-state.coffee:192:45)
  at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/Applications/Atom.app/Contents/Resources/app.asar/src/command-registry.js:238:29)
  at /Applications/Atom.app/Contents/Resources/app.asar/src/command-registry.js:3:61
  at KeymapManager.module.exports.KeymapManager.dispatchCommandEvent (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/atom-keymap/lib/keymap-manager.js:519:16)
  at KeymapManager.module.exports.KeymapManager.handleKeyboardEvent (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/atom-keymap/lib/keymap-manager.js:354:22)
  at HTMLDocument.module.exports.WindowEventHandler.onKeydown (/Applications/Atom.app/Contents/Resources/app.asar/src/window-event-handler.js:178:20)

```

### Commands

```
  2x -0:21.4.0 blur (atom-text-editor.editor.mini.is-focused)
[...]

PR #715, opened 2015-06-25
fix Uncaught NotFoundError

jacekkopecky:
The error was caused by event handlers of `failed-to-compose` cancelling the operation that’s already being cancelled.

Replicate the issue by pressing `cm`.

May be what caused #713, #646, #642 - and certainly is the one described in #633.


jacekkopecky:
I just hit the 1.0.0 release rush and Travis couldn't get Atom - any way to try the build again when the seas are calmer?

maxbrunsfeld:
:dragon: Awesome. Thanks so much for fixing this @jacekkopecky.

PR #718, opened 2015-06-25
make find state global

jacekkopecky:
Makes find state global, like search state (history) is shared across text editors.
An example: if you do `t*` to move the cursor to the next `*` in one editor, switch to another editor, and do `;`, you'll go to the next `*` as well.

### DIFF
--- a/spec/motions-spec.coffee
+++ b/spec/motions-spec.coffee
@@ -19,7 +19,8 @@ describe "Motions", ->
     helpers.keydown(key, options)
 
   normalModeInputKeydown = (key, opts = {}) ->
-    editor.normalModeInputView.editorElement.getModel().setText(key)
+    theEditor = opts.editor or editor
+    theEditor.normalModeInputView.editorElement.getModel().setText(key)
 
   submitNormalModeInputText = (text) ->
     inputEditor = editor.normalModeInputView.editorElement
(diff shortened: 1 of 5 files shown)


bronson:
Love it!  This is how Vim works.

spec/motions-spec.coffee
```diff
@@ -1735,6 +1735,25 @@ describe "Motions", ->
       keydown(',')
       expect(editor.getCursorScreenPosition()).toEqual [0, 2]
 
+    it "gets state from GlobalVimState", ->
+      keydown('f')
+      normalModeInputKeydown('c')
+      expect(editor.getCursorScreenPosition()).toEqual [0, 2]
+
+      # store the current find for 'c'
+      findFC = vimState.globalVimState.currentFind
```
maxbrunsfeld: Instead of accessing `vimState.globalVimState.currentFind` in this test, could we create a second text editor instance and run a find command in it?

I think that something like this should work:

```coffee
helpers.getEditorElement (otherEditorElement) ->
  otherEditorElement.getModel().setText("abc")
  keydown('f', element: otherEditorElement)
  keydown('c', element: otherEditorElement)

  keydown(';')
  expect(editor.getCursorScreenPosition()).toEqual [0, 8]
```

maxbrunsfeld: Our test helpers are a bit awkward right now for creating other editors. I'd like to simplify them.

jacekkopecky:
tests updated to actually use multiple editors (thanks for the suggestion @maxbrunsfeld); please review especially the changes in spec-helper - hope this is useful for other specs too.

spec/spec-helper.coffee
```diff
@@ -17,13 +18,18 @@ getEditorElement = (callback) ->
...
+    element.vimState = new VimState(element, statusBarManager, globalVimState)
 
     element.addEventListener "keydown", (e) ->
       atom.keymaps.handleKeyboardEvent(e)
 
     callback(element)
 
+# this function puts the existingElement parameter first
+getOtherEditorElement = (existingElement, callback) -> getEditorElement(callback, existingElement)
+
```
maxbrunsfeld: I think the best way of sharing the `GlobalVimState` between the two editors would be to store it as a local variable in this file (`globalVimState = null` at the top of the file) and initialize it in this file's `beforeEach` block. That way, we could create other editors without needing a reference to the first editor.

jacekkopecky: very good point, how about now? 8-)

spec/spec-helper.coffee
```diff
@@ -3,10 +3,15 @@ GlobalVimState = require '../lib/global-vim-state'
...
+[globalVimState, statusBarManager] = []
+
 beforeEach ->
   atom.workspace ||= {}
+  statusBarManager = null
+  globalVimState = null
 
-getEditorElement = (callback) ->
+getEditorElement = (callback, existingElement = {}) ->
+  existingVimState = existingElement?.vimState
```
maxbrunsfeld: :fire: existingVimState?

spec/motions-spec.coffee
```diff
@@ -1744,6 +1745,38 @@ describe "Motions", ->
       keydown(',')
       expect(editor.getCursorScreenPosition()).toEqual [0, 2]
 
+    it "gets state from GlobalVimState", ->
```
maxbrunsfeld: I think you can avoid making reference to the implementation here: something like `it "shares the most recent find command with other editors", ->`

jacekkopecky: yup, done too. All good catches, thank you. I also changed a find to a till to test both.

spec/spec-helper.coffee
jacekkopecky: oh boy, almost missed this one, fixed too

jacekkopecky: fixed, take two, sorry

maxbrunsfeld:
Awesome!

jacekkopecky:
thanks for the patience and attention to detail 8-)

PR #720, opened 2015-06-26
Insert mode put

jacekkopecky:
In insert mode in VIM, you can press `ctrl-r` followed by a register name to paste the contents of the register.


bronson:
Very cool.

I agree, the ctrl-e conflict is worrisome.  I really like that normal mode feels like Vim and insert mode feels like Atom.

Would it make sense to give vim-mode a configuration option: "Vim-Flavored Insert Mode" and "Atom-Flavored Insert Mode"?  When there's a conflict, this setting would tell whose keybindings to prefer.

I'd set mine to Atom-flavored.

maxbrunsfeld:
This is really well done. My apologies for not seeing this sooner.

jacekkopecky:
Thanks, and no worries about the timing, I understand that getting Atom 1.0 out the door made everyone quite busy. 8-)

PR #721, opened 2015-06-26
feature: copy character from line above/below

jacekkopecky:
This PR implements VIM insert-mode commands `ctrl-y` and `ctrl-e` for copying from the line above/below the cursor.
A few questions in relation to this PR:
 * is `lib/insert-mode.coffee` an OK new file for this? 
 * if so, should #720 live there, too?
 * should I worry that some might use `ctrl-e` for going to the end of the line? (default in Atom)

Oh, and congratulations on Atom 1.0.0! :fireworks: :balloon:

### DIFF
--- a/spec/insert-mode-spec.coffee
+++ b/spec/insert-mode-spec.coffee
@@ -1,0 +1,68 @@
+helpers = require './spec-helper'
+
+describe "Insert mode commands", ->
+  [editor, editorElement, vimState] = []
+
+  beforeEach ->
+    vimMode = atom.packages.loadPackage('vim-mode')
+    vimMode.activateResources()
+
+    helpers.getEditorElement (element) ->
+      editorElement = element
+      editor = editorElement.getModel()
+      vimState = editorElement.vimState
+      vimState.activateNormalMode()
+      vimState.resetNormalMode()
+
+  keydown = (key, options={}) ->
+    options.element ?= editorElement
+    helpers.keydown(key, options)
+
+  describe "Copy from line above/below", ->
+    beforeEach ->
+      editor.setText("12345\n\nabcd\nefghi")
+      editor.setCursorBufferPosition([1, 0])
+      editor.addCursorAtBufferPosition([3, 0])
+      keydown 'i'
+
+    describe "the ctrl-y command", ->
+      it "copies from the line above", ->
+        keydown 'y', ctrl: true
+        expect(editor.getText()).toBe '12345\n1\nabcd\naefghi'
+
+        editor.insertText ' '
+        keydown 'y', ctrl: true
+        expect(editor.getText()).toBe '12345\n1 3\nabcd\na cefghi'
+
+      it "does nothing if there's nothing above the cursor", ->
+        editor.insertText 'fill'
+        keydown 'y', ctrl: true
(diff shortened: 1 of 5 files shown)


bronson:
Very cool.

I agree, the ctrl-e conflict is worrisome.  I really like that normal mode feels like Vim and insert mode feels like Atom.

Would it make sense to give vim-mode a configuration option: "Vim-Flavored Insert Mode" and "Atom-Flavored Insert Mode"?  When there's a conflict, this setting would tell whose keybindings to prefer.

I'd set mine to Atom-flavored.

bronson:
Oh heck, users can always bind ctrl-e back to editor:move-to-end-of-line if they care.  At least this way there's a choice.

Merged it to vim-mode-next.  This PR looks great.

bronson:
Getting strange errors when running this patch in vim-mode-next: https://travis-ci.org/bronson/vim-mode-next/builds/69569825

Looks related to the activateResources/getEditorElement mocking...  Maybe a timing issue?

To reproduce locally:

    git clone https://github.com/bronson/vim-mode-next
    git checkout 0ad53fb
    cd vim-mode-next
    apm install
    apm test

Very curious what's going on.

Since your patch works fine in vim-mode, feel free to ignore.  :)

jacekkopecky:
just guessing, but it seems like it needed a change of `'vim-mode'` to `'vim-mode-next'` in the spec file: https://github.com/bronson/vim-mode-next/blob/0ad53fbfa4b2c626e00d7e16c98e95fc6232e920/spec/insert-mode-spec.coffee#L7

bronson:
DOH!!  Like this?  https://github.com/bronson/vim-mode-next/commit/654a9c3a97eceba08aaa19f21f6218ff58be9c3e

You're right, how did I miss that?  Sorry.  :)

bronson:
Yep, that was it.  vim-mode-next 0.54.11 has ctrl-e/ctrl-y support again.

maxbrunsfeld:
I think I'd like to *not* override the `ctrl-e` key-binding by default. I agree with @bronson that by default, insert mode should feel like Atom. It's great to have the command implemented though: anybody who wants this command can just add the keybinding to their `keymap.cson`.

jacekkopecky:
No worries, done. I tend to use `ctrl-y` a lot, but I'm not sure I've ever used `ctrl-e` anyway.

spec/insert-mode-spec.coffee
```diff
@@ -0,0 +1,73 @@
...
+        editor.insertText 'a'
+        expect(editor.getText()).toBe '12a345\n\nabcd\nefaghi'
+        keydown 'y', ctrl: true
+        expect(editor.getText()).toBe '12a345\n\nabcd\nefadghi'
+
+    describe "the ctrl-e command", ->
+      beforeEach ->
+        atom.keymaps.add "test",
+          'atom-text-editor.vim-mode.insert-mode':
+            'ctrl-e': 'vim-mode:copy-from-line-below'
```
maxbrunsfeld: :smile:

jacekkopecky: well we're not throwing away perfectly good specs, are we? 8-)

PR #727, opened 2015-06-29
Implement gf (goto file)

irrationalistic:
This implements the gf keybinding which will allow the editor to open the file or url under the cursor. It supports absolute file paths, relative file paths, and even paths without an extension. If no extension is found, but multiple files are found with the same name (i.e. main.js or main.coffee), it will look to the currently loaded file for advice on the extension to use.

I've run into an issue where I can't seem to get tests to work properly. The situation is that in order for relative paths to work, there needs to be an actual file (from the file system) loaded into the editor. Any suggestions on how to get that working? I'd also need to be able to test that once the command is entered new files open as well.

### DIFF
--- a/lib/motions/open-file-under-cursor.coffee
+++ b/lib/motions/open-file-under-cursor.coffee
@@ -3,6 +3,7 @@ path = require 'path'
 {Directory} = require 'atom'
 shell = require 'shell'
 {exec} = require 'child_process'
+utils = require '../utils'
 
 class OpenFileUnderCursor extends Motion
   execute: (count) ->
@@ -17,8 +18,11 @@ class OpenFileUnderCursor extends Motion
     ###
     wordRegex = /[a-z0-9\.\-_\/\\%:]+/i
     @editor.getCursors().forEach (cursor) =>
-      range = cursor.getCurrentWordBufferRange wordRegex: wordRegex
-      selectedPath = @editor.getTextInRange range
+      currentWord = utils.getCurrentWord @editor, cursor, wordRegex
+      # range = cursor.getCurrentWordBufferRange wordRegex: wordRegex
+      # selectedPath = @editor.getTextInRange range
+      range = currentWord.range
+      selectedPath = currentWord.word
 
       # exit early for url match
       if /^https?:/i.test(selectedPath)
--- a/lib/motions/open-file-under-cursor.coffee
+++ b/lib/motions/open-file-under-cursor.coffee
@@ -1,0 +1,54 @@
+{Motion} = require './general-motions'
+path = require 'path'
+{Directory} = require 'atom'
+shell = require 'shell'
+{exec} = require 'child_process'
+
+class OpenFileUnderCursor extends Motion
+  execute: (count) ->
+    ###
+      /usr/bin/test
+      ../../.travis.yml
+      ./search-motion
+      ./index
+      http://github.com
+      https://thing.com/item%20test
+      //github.com
+    ###
+    wordRegex = /[a-z0-9\.\-_\/\\%:]+/i
+    for cursor in @editor.getCursors()
+      range = cursor.getCurrentWordBufferRange wordRegex: wordRegex
+      selectedPath = @editor.getTextInRange range
+
+      # exit early for url match
+      if /^https?:/i.test(selectedPath)
+        return @openUrl selectedPath
+      if /^\/\//i.test(selectedPath)
+        return @openUrl "http:#{selectedPath}"
+
+      if selectedPath[0] isnt '/' and selectedPath[0] isnt '\\'
+        selectedPath = path.join path.dirname(@editor.buffer.file.path), selectedPath
+
+      pathDir = path.dirname selectedPath
+      dir = new Directory pathDir
+      ext = path.extname @editor.buffer.file.path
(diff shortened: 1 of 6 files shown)


masukomi:
Can't comment on the testing but the PR could be improved by rebasing all the commits into 1. Or maybe 2: one for the initial implementation, and one for implementing multiple file selections.

bronson:
Hi, looks good.

Any chance of reusing [getCurrentWord](https://github.com/atom/vim-mode/blob/master/lib/motions/search-motion.coffee#L94-L110) instead of re-implementing it here?

That might require some refactoring...  but it might be worth it since Vim seems to need the current word quite a bit.

irrationalistic:
I imagine that would be doable. Probably move it into a base class or utilities somewhere? I'll take a look.

irrationalistic:
I've refactored it into the utils file. It can be used from any component now and supports any cursor(s) and word regular expressions!

bronson:
Looks excellent!!  Now it just needs some testing.  :)

I can try to help with that if you want.  Got a few things I need to tie down first though.

irrationalistic:
That'd be awesome, thanks! I tried to get some stuff working, but no go. The problem is that it needs to have an actual file loaded since the relative path testing needs to be relative to the loaded file. Any thoughts you have would be great!

lib/motions/open-file-under-cursor.coffee
```diff
@@ -0,0 +1,58 @@
...
+  execute: (count) ->
+    ###
+      /usr/bin/test
+      ../../.travis.yml
+      ./search-motion
+      ./index
+      http://github.com
+      https://thing.com/item%20test
+      //github.com
+    ###
```
bronson: I suppose this comment will become tests once it's clear how to write them

lib/motions/open-file-under-cursor.coffee
```diff
@@ -0,0 +1,58 @@
...
+    wordRegex = /[a-z0-9\.\-_\/\\%:]+/i
+    @editor.getCursors().forEach (cursor) =>
+      currentWord = utils.getCurrentWord @editor, cursor, wordRegex
+      # range = cursor.getCurrentWordBufferRange wordRegex: wordRegex
+      # selectedPath = @editor.getTextInRange range
+      range = currentWord.range
+      selectedPath = currentWord.word
```
bronson: Would `{range, selectedPath} = utils.getCurrentWord(...)` work?

lib/motions/open-file-under-cursor.coffee
```diff
@@ -0,0 +1,58 @@
...
+
+      # exit early for url match
+      if /^https?:/i.test(selectedPath)
+        return @openUrl selectedPath
+      if /^\/\//i.test(selectedPath)
+        return @openUrl "http:#{selectedPath}"
+
+      if selectedPath[0] isnt '/' and selectedPath[0] isnt '\\'
+        selectedPath = path.join path.dirname(@editor.buffer.file.path), selectedPath
+
```
bronson: maybe put everything below here into a new function?  overall this patch looks great to me.

lib/motions/open-file-under-cursor.coffee
irrationalistic: I wanted to be able to use selectedPath, but keep the `getCurrentWord` method as generic as possible, so i'd end up with an extra declaration anyways or I'd have to use `word` in the remaining code! Does that make sense?

lib/motions/open-file-under-cursor.coffee
irrationalistic: Thanks! Not sure if that would really be useful yet, but it might be good if there is a need to extend from this class. Are there any other vim commands that work via files?

lib/motions/open-file-under-cursor.coffee
bronson: oop, you're right.  renaming word to selectedPath takes another line anyway so this change wouldn't buy much.  Never mind!

lib/motions/open-file-under-cursor.coffee
bronson: There are some ex functions that work on files like this but I can't think of anything else that would be in vim-mode.  So, yea, doesn't make sense from a code reusability standpoint.

I was just thinking from a readability standpoint: seems like the first half of the function is doing a pretty different job from the second half.  The parts might be easier to understand than the whole.

But, like before, I might not be seeing the whole picture here.  :)

irrationalistic: You are totally right about improving readability! I can take a look at what the refactor would be like.

irrationalistic: So I realized that there are references to `openUrl` on 29 and 30, which are about, technically, opening files as well. A refactor probably wouldn't make a lot of sense at this point.

t9md:
Fore reference, I also implemented similar feature with [open-this](https://atom.io/packages/open-this), the diff of this PR is.

1. Try every possible ext name using grammar's FileTypes array(it include ext name list per language).
2. Try to open the file with same base name if it can't find appropriate file in 1.

https://github.com/t9md/atom-open-this/blob/master/lib/main.coffee#L18-L24

And not include opening URL.

lee-dohm:

PR #732, opened 2015-07-01
Make control-t return from symbol, not control-shift-t

bronson:
This updates vim-mode to match Vim's binding.

I assume this was a simple typo?  https://github.com/atom/vim-mode/commit/c35d2d6744c3beaa91f8bc9e82ee19f11680478b

### DIFF
--- a/keymaps/vim-mode.cson
+++ b/keymaps/vim-mode.cson
@@ -162,7 +162,7 @@
   '.': 'vim-mode:repeat'
 
   'ctrl-]': 'symbols-view:go-to-declaration'
-  'ctrl-T': 'symbols-view:return-from-declaration'
+  'ctrl-t': 'symbols-view:return-from-declaration'
 
   'ctrl-a': 'vim-mode:increase'
   'ctrl-x': 'vim-mode:decrease'


maxbrunsfeld:
Thanks!

tmm1:
There was some concern that this would affect windows users where CTRL-t is used for an existing atom feature

maxbrunsfeld:
That makes sense. Probably [toggling the fuzzy-finder](https://github.com/atom/fuzzy-finder/blob/master/keymaps/fuzzy-finder.cson#L8). While there is already an alternative keybinding for that command: `ctrl-p`, I think it's probably best not to override key-bindings that ship with Atom.

In my opinion, we should simply remove this `return-from-declaration` keybinding on linux and windows, and let users add it (or some modified version like `ctrl-shift-T`) themselves. @bronson @tmm1 thoughts?

bronson:
Thanks @tmm1, that makes sense.

Since the user asked for vim-mode, wouldn't she expect ctrl-p to be fuzzy-finder and ctrl-t to be return-from-tag?

And, if vim-mode's normal mode doesn't get to override ctrl-t, then why does it get to override ctrl-u, ctrl-b, ctrl-d, ctrl-f, ctrl-e, etc?

I _definitely_ agree with not stomping on Atom keybindings when in insert mode!

maxbrunsfeld:
> And, if vim-mode's normal mode doesn't get to override ctrl-t, then why does it get to override ctrl-u, ctrl-b, ctrl-d, ctrl-f, ctrl-e, etc?

That's a fair point, since this is only in normal mode, it's not such a big deal to override another binding.

PR #738, opened 2015-07-04
make tests fail if deprecated calls are made

bronson:
Happy to say, master doesn't call any deprecated APIs (while testing anyway).

But this might be needed soon...  there are some pull requests that do.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     }
   },
   "devDependencies": {
-    "coffeelint": "^1.10.1"
+    "coffeelint": "^1.10.1",
+    "grim": "^1.2.1"
   }
 }
--- a/spec/spec-helper.coffee
+++ b/spec/spec-helper.coffee
@@ -2,9 +2,17 @@ VimState = require '../lib/vim-state'
 GlobalVimState = require '../lib/global-vim-state'
 VimMode  = require '../lib/vim-mode'
 StatusBarManager = require '../lib/status-bar-manager'
+Grim = require 'grim'
 
 beforeEach ->
   atom.workspace ||= {}
+
+afterEach ->
+  if Grim.getDeprecationsLength() > 0
+    Grim.logDeprecations()
+    Grim.clearDeprecations()
+    this.fail 'Deprecated APIs were called'
+
 
 getEditorElement = (callback) ->
   textEditor = null


jacekkopecky:
I'm getting 
```
      ReferenceError: fail is not defined
        at [object Object].<anonymous> (/Users/jacek/prog/oss/vim-mode/spec/spec-helper.coffee:13:5)
```
Is it just my version of jasmine being somehow outdated? If so, how do I update the version of jasmine that is used by `apm test`?

bronson:
Wow!  And you ran `apm install` again?  I'll look...

bronson:
It looks like fail() was added to Jasmine 2.1: https://github.com/jasmine/jasmine/commit/b1344d5c73c5e01a07e1ea435be3ed980f6db9de

jacekkopecky:
Yes, but my quick search just now seems to show that `apm` testing still uses jasmine 1.3 (https://github.com/atom/atom/pull/6099 and https://github.com/atom/atom/pull/5393). Did you do anything to upgrade that? Or am I missing something?

bronson:
Agreed!  No, I haven't done anything that I'm aware of, and the continuous build passed.  Do your version numbers match?

```
$ apm --version
apm  1.0.1
npm  2.5.1
node 0.10.35
python 2.7.9
git 2.2.1
$ atom --version
1.0.0
```

By setting breakpoints, I'm certain that I'm using the vendor/jasmine.js file from Atom's app.asar.  And, at line 2325 there's this:

```js
jasmine.Spec.prototype.fail = function (e) {
  var expectationResult = new jasmine.ExpectationResult({
    passed: false,
    message: e ? jasmine.util.formatException(e) : 'Exception',
    trace: { stack: e.stack }
  });
  this.results_.addResult(expectationResult);
};
```

So I'm as baffled as anyone, both why it fails and why it works.

jacekkopecky:
I now found that `this.fail` works while simple `fail` doesn't, on my machine. I see the same code you see.

The failing `fail` will still fail the test because of that `ReferenceError`, and the logs will still show the right deprecation call, so that's not a problem.

Also, when I merged #738 into my own private version, `apm test` would fail every test after the first one that used a deprecated call. 

So I ended up with this patch against #738:
```
--- a/spec/spec-helper.coffee
+++ b/spec/spec-helper.coffee
@@ -10,7 +10,8 @@ beforeEach ->
 afterEach ->
   if Grim.getDeprecationsLength() > 0
     Grim.logDeprecations()
+    Grim.clearDeprecations()
-    fail 'Deprecated APIs were called'
+    this.fail 'Deprecated APIs were called'
 
 
 getEditorElement = (callback) ->
```

bronson:
Well at least all that makes sense.  Updated, thanks!

bronson:
Also updated packages.json  to specify grim version `"^1.2.1"` instead of `"1.4.1"`

maxbrunsfeld:
Sorry for weighing in so late on this. I don't think tests should *fail* because of deprecations, because the whole idea of deprecations is that we want to remove some functionality, but we want to give packages some time to upgrade before we break them.

In general, to me, a CI build should only be red on the master branch if the authors have made a *mistake*. That way, a red build can always be treated to mean "fix this before pushing any other code changes". If we make the build go red for lower-priority problems like use of deprecated APIs, then red builds carry less weight.

bronson:
I get that.  It would be trivial to change it to print warnings into the Travis logs.  That would make the problem visible before actually using the feature.  Care to have that change?

(A better technique might be to print the warnings from the call sites, rather than collecting and printing them at the end of the test, but I'd have to learn how that code worked first...  Tell me if you'd prefer that.)

maxbrunsfeld:
That's a good point: we only surface the deprecations when running the specs in Atom, not at the command line. I think that's something we should fix in Atom core. Could you open an issue about this? If you're interested in fixing it, I think we'd need to do the same [thing that we do in Atom's graphical spec reporter](https://github.com/atom/atom/blob/8586a6f2dcd30dfb4c6595c46e529c819afb7829/spec/atom-reporter.coffee#L223), when we [run the tests in the terminal](https://github.com/atom/atom/blob/8586a6f2dcd30dfb4c6595c46e529c819afb7829/spec/jasmine-helper.coffee#L26).

maxbrunsfeld:
OOH, I just realized that there is *already* code for doing that, but it only happens on our internal CI system, Janky. We could just change that `if process.env.JANKY_SHA1` to say `if process.env.JANKY_SHA1 or process.env.TRAVIS` or whatever environment variables there are on travis.

lee-dohm:

Issue #246, opened 2014-05-15
Repetitive visual-mode block operations


opichals:
Steps in vim:
- Mark a few lines using visual mode
- Indent using '>'
- Indent the whole block again using '.'

Atom appears to forget the block and rather just indents the cursor line at the moment

Atom Version 0.95.0 (0.95.0)

maxbrunsfeld:
I think this works now. Let me know if you still have problems with it.

techniq:
@maxbrunsfeld - Doing the steps in the original report are still an issue (make a visual selection, indent, then press `.` to indent again).  If I escape visual mode after indenting, pressing `.` to repeat the indention works.

maxbrunsfeld:
@techniq thanks, I see. So currently, `.` is not bound in visual mode. Do you all think that the proper solution would be to add a `.` binding in visual mode?

techniq:
@maxbrunsfeld - I think adding the binding would be right, but others might be better to chime in.

james147:
Note that Vim's default behaviour for visual indent is to exit visual mode
straight away and . then repeats the indent for the number of lines that
where selected starting at the cursor.

techniq:
I think adhering to Vim's default behavior would be best.  It's been a while since I've used Vim directly (used Sublime with vintage for a while before moving over to Atom).  Although staying in visual mode would allow repeated indenting with `>>` (or back with `<<`), instead of the initial `>>` or `<<` and then repeating the indention with `.`.  If you over indented with repeat, you'd have to restore your selection, and it currenlty looks like `gv` doesn't work with vim-mode to restore the last selection.

maxbrunsfeld:
Yeah, I actually like the fact that it stays in visual mode right now; it's kind of convenient. That said, I'm open to changing it to be more like vim if it's the only way to make the repeating use case work.

techniq:
If adding the `.` binding in visual mode works for others, I say that would
be the way to go.

masukomi:
Personally i *hate* vim's default behavior in this case. I had to go and figure out how to make it stop exiting visual mode every time I indented because it was driving me nuts. So, while I agree with folks about making vim's default behavior the default it'd be really, *really* nice is you had the option to stay in visual mode after performing an operation in it.

bronson:
I agree 100%.  I think Vim got this wrong.  Would love to see vim-mode stay in visual mode.

flannelhead:
[...]

PR #747, opened 2015-07-08
Exit visual mode on indent/unindent

bronson:
First stab at fixing #246

### DIFF
--- a/lib/operators/indent-operators.coffee
+++ b/lib/operators/indent-operators.coffee
@@ -8,10 +8,9 @@ class AdjustIndentation extends Operator
 
     @indent()
 
-    if mode isnt 'visual'
-      @editor.setCursorBufferPosition([start.row, 0])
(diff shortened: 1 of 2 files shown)


bronson:
In vim-mode-next 0.54.15.  It appears to work great, including repeating.

There's one difference from Vim: the cursor moves to the beginning of the line instead of remaining where it is.  I can open another issue to address that if desired.

maxbrunsfeld:
Thanks for doing this. 

> the cursor moves to the beginning of the line instead of remaining where it is. I can open another issue to address that if desired

I also noticed that the existing code will not preserve multiple selections/cursors; it destroys all but one cursor afterward by calling `::setSelectedBufferRange()`. It'd be nice to fix both of these things in separate PRs. In the meantime, this looks good. Just waiting on Travis.

maxbrunsfeld:
:tada:

jacekkopecky:
Yay!

PR #757, opened 2015-07-09
implement substitute using the change operator

bronson:
Removes some code duplication.  Counts still work of course.

There might be some other places we can do this but not sure if there's interest there is in this sort of cleanup.


jacekkopecky:
Please rebase. 8-)

bronson:
Doh, right!  Rebased.

maxbrunsfeld:
Wow nice.

PR #762, opened 2015-07-11
Implement ex mode

jazzpi:
The reasons why implementing `ex` mode in `vim-mode` itself is a good idea are pretty well laid out in #596.

As a follow-up, this PR attempts to implement `ex`. It's definitely not done yet (see the to-do list), but I would like to get some early feedback since this is a pretty big chunk of code (and it is still not quite clear to me whether or not this is actually wanted).

So far, a lot of the code was simply copy-pasted over from [ex-mode](https://github.com/lloeki/ex-mode), however there is one crucial difference:

Ex commands are not added by simply adding a method to the `Ex` class. Instead, they are added by calling `ExCommand.registerCommand`. The reasons why the behemoth-class way didn't work are discussed in lloeki/ex-mode#49. They are also passed an object instead of an argument list.

This allows for a simplified method of matching methods - Since every command has an associated priority with it, if multiple commands match an input (e.g. `:e` matching both `:edit` and `:echo`), the one with the higher priority is executed. If multiple commands with the same priority match the input, the one that comes first in the alphabet is executed.

@romgrk proposed a different format in lloeki/ex-mode#49, but I felt like this format requires less effort and is less prone to errors. I would however like some feedback on this topic.

Now, here is a Todo list of things necessary before merging this:

 - [x] Don't use SearchViewModel, so <kbd>:</kbd> and <kbd>/</kbd> have their own histories.
 - [x] Implement the rest of the commands [ex-mode](https://github.com/lloeki/ex-mode) currently supports for a seemless transition
 - [x] Specs!
 - [x] `:/a/d` style commands should work.
 - [x] The `:/` ~~command~~ range should work.

Some things that would be nice to have:
 - [ ] Add support for `[count]tabnext` style commands
 - [ ] Tab completion
 - [ ] Support for commands like `:'<,'>s/a/b` (Visual mode ranges)
 - [ ] Setting Atom/vim-mode options with `:set`
 - [ ] More commands, obviously
 - [x] Expose `registerCommands` via a service to make adding commands from `init.coffee` possible

The last one is a bit obscure, so here are some examples:

```coffeescript
atom.packages.onDidActivatePackage (pack) ->
  if pack.name is 'vim-mode'
    commands = pack.mainModule.provideVimMode().getGlobalState().exCommands
    commands.registerCommand(
      name: 'z'
      priority: 1
      callback: ->
        console.debug('zzz'))
    commands.registerCommand(
      name: 'test'
      priority: 1
      callback: ({range, args, editor}) ->
        range = [[range[0], 0], [range[1] + 1, 0]]
        editor.getBuffer().setTextInRange(range, args)
    )
```


lloeki:
Thanks @jazzpi.

jazzpi:
Alright, this should be finished, finally :smile:

ardrigh:
Thanks for the work on this.

Pardon my ignorance, but what do I need to do to help test this as a package in Atom?

jazzpi:
Thanks for helping out :smile:

What you'll want to do is opening a terminal and executing these commands:

```bash
git clone https://github.com/jazzpi/vim-mode
cd vim-mode
git checkout ex-mode
apm link -d
```

This will download the code, check out into the feature branch and then create a symlink in your dev packages so that when you start Atom in development mode (`atom -d`), it will use this version instead of the version published to atom.io

If you have the `ex-mode` package installed, you'll also want to disable that. After you're done testing, go to the directory where you downloaded the code to and execute `apm unlink -d` to remove  the symlink.

ardrigh:
I tried the instructions to get the package, but it failed to load. I disabled my other 3rd party packages to try running this.

I am using Windows 7, Atom 1.0.2

Download of git repo was OK.
Link was set

apm link -d
C:\Users\<user>\.atom\dev\packages\vim-mode -> c:\Local\git\vim-mode

Started with 'atom -d'

Failed to load the vim-mode package
Cannot find module 'event-kit'
Hide Stack Trace
Error: Cannot find module 'event-kit'
    at Module._resolveFilename (module.js:328:15)
    at Function.Module._resolveFilename (C:\Users\<user>\AppData\Local\atom\app-1.0.2\resources\app.asar\src\module-cache.js:383:52)
    at Function.Module._load (module.js:270:25)
    at Module.require (module.js:357:17)
    at require (module.js:376:17)
    at Object.<anonymous> (c:\Local\git\vim-mode\lib\vim-mode.coffee:4:10)
    at Object.<anonymous> (c:\Local\git\vim-mode\lib\vim-mode.coffee:70:4)
    at Module._compile (module.js:452:26)
    at Object.requireCoffeeScript (C:\Users\<user>\AppData\Local\atom\app-1.0.2\resources\app.asar\node_modules\coffee-cash\lib\coffee-cash.js:85:19)
    at Module.load (module.js:347:32)

Any ideas?

jazzpi:
Try running `apm install .` in the `vim-mode` directory.

Also, did the `git checkout` run without errors? I just realized the command should be `git checkout origin/ex-mode` :flushed:

jazzpi:
This isn't going to get merged, see https://github.com/atom/vim-mode/issues/596#issuecomment-125678155 for the reasons.

Issue #680, opened 2015-05-25
`yi/` opens search


jazzpi:
I tried yanking the contents of a RegEx using `yi/`, which opened search. After that I checked vim, which didn't open search but also didn't yank the text enclosed by `/` (that would be great).

bronson:
So, you're asking for a // text object?

I agree that, if Atom doesn't recognize a text object, it should probably do nothing.  Like vim.

jazzpi:
Yes, it should probably do nothing - and not open search.

lee-dohm:
As stated in [the README](https://github.com/atom/vim-mode), this package is no longer maintained and is deprecated. We recommend that people use the [vim-mode-plus package](https://atom.io/packages/vim-mode-plus) instead. Because of this, we are [archiving this repository](https://help.github.com/articles/about-archiving-repositories/) and closing all issues and pull requests. Thanks very much for your support and contributions!

Issue #657, opened 2015-05-08
Operators replace text with motion when used with pane splits


jazzpi:
If you split a pane and then delete a line in the original pane using `dd`, the line gets deleted but a `d` is inserted at the beginning of the next line (even after closing the opened pane).

The same thing happens with `cc` to change the line (except it inserting a `c`, not a `d`).

jazzpi:
I just noticed this seems to apply to any operator with motions - for example, going to the beginning of the document and then pressing `G` to yank the entire document does that - and replace the document with `G`. This basically makes pane splits unusable.

bronson:
That's strange, I can't reproduce this.  I split a pane (both using ex mode's :split and the menu), hit dd, and both panes had the correct text.  Tried a few other motions too.

Does it happen when you only have vim-mode enabled, no other 3rd party plugins?

jazzpi:
I think this fixed itself when I upgraded Atom, I'm unable to reproduce this now.

romgrk:
Still poping for me, with atom v0.204.0.
Happens only when I'm working with split panes.

jazzpi:
Strange, it happens again for me now. I don't know why, but I'm pretty sure it worked for a while - maybe I just worked in the new pane everytime?

jazzpi:
Alright, here's something interesting:

* Open a new window
* Enter
```
abc
def
ghi
```
into the `untitled` file
* Press `Ctrl-V`, then `Ctrl-H`
* Go to the first line, press `dd`. File now looks like
```
ddef
ghi
```
* Press `u`, then press `cc`, then `Esc`. File now looks like
```
abc
c
ghi
```
* Press `u`, then delete a line. It now works correctly again.

jazzpi:
This happens too when I disable all packages but vim-mode.
If I do this on a clean install with only vim-mode installed, it works however.

ssfrr:
I'm seeing this as well, and actually originally attributed it to bronson/vim-mode-next#1. I can reliably trigger it by opening a vertical split, or closing a vertical split if I opened Atom with one. Closing the split again doesn't fix the problem. After I restart Atom the problem goes away though. I tried this with all other plugins disabled. It does seem to be reliably reproducible.

Strangely doing `yy` to yank a line causes the line to disappear (and the letter `y` is prepended to the next line).

Yanking from the command palette does not delete the line though.

bronson:
[...]

PR #764, opened 2015-07-11
cancel operator-pending on unrecognized key strokes

### LINKED ISSUE #734: if text object isn't recognized, nothing should happen
If the user types `vix`, nothing should happen.   `x` is not a valid text object.

Vim does nothing.  Atom, alas, deletes the character under the cursor.

Inspired by https://github.com/atom/vim-mode/issues/680

jacekkopecky:
_(edited)_ Fix #734 and #680. This is an alternative to #781, because of issues noted below in https://github.com/atom/vim-mode/pull/764#issuecomment-122441219 and further comments. Discussion welcome. 

This fix overrides Atom's normal key binding resolution process for multi-keystroke key bindings: where Atom will do `ci3w` like this:
1. `c` -> change
2. `i` -> _some potential bindings, e.g. `iw`, `ip` etc., so let's wait_
3. `3` -> `i3` is not known and no key binding starts with `i3` so let's break it down
   1. `i` on its own is unknown in operator-pending mode, so let's ignore it
   2. let's try again `3` -> a prefix, good
4. `w` combines with `3` so the final result is `c3w` 

What we want is
1. `c` -> change
2. `i` as above
3. `3` leads to `i3` which is unknown and cancels `c`
4. `w` moves cursor to next word

We could beep on a cancellation like this.

The `atom.keymaps.cancelPendingState()` call is a private API, but I can't judge whether that's too bad for this PR.

### DIFF
--- a/lib/utils.coffee
+++ b/lib/utils.coffee
@@ -1,3 +1,6 @@
+# copied from atom/atom-keymap src/helpers.coffee
+AtomModifierRegex = /(ctrl|alt|shift|cmd)$/
+
 module.exports =
   # Public: Determines if a string should be considered linewise or character
   #
@@ -12,3 +15,7 @@ module.exports =
       'linewise'
     else
       'character'
+
+  # copied and simplified from atom/atom-keymap src/helpers.coffee
+  isAtomModifier: (keystroke) ->
+    AtomModifierRegex.test(keystroke)
--- a/lib/vim-state.coffee
+++ b/lib/vim-state.coffee
@@ -36,6 +36,12 @@ class VimState
         @activateVisualMode('characterwise') if @mode is 'command'
     , 100)
 
+    @subscriptions.add atom.keymaps.onDidFailToMatchBinding (e) =>
+      return unless e.keyboardEventTarget is @editorElement
+      if @mode is 'operator-pending'
+        atom.keymaps.cancelPendingState()
+        @resetCommandMode()
+
     @editorElement.classList.add("vim-mode")
     @setupCommandMode()
     if settings.startInInsertMode()
--- a/lib/vim-state.coffee
+++ b/lib/vim-state.coffee
@@ -39,10 +39,15 @@ class VimState
     @subscriptions.add atom.keymaps.onDidFailToMatchBinding (e) =>
       return unless e.keyboardEventTarget is @editorElement
       return if Utils.isAtomModifier(e.keystrokes)
-      
-      atom.keymaps.cancelPendingState()
+
       if @mode is 'operator-pending'
+        atom.keymaps.cancelPendingState()
+        atom.beep()
         @resetCommandMode()
+
+      if @mode is 'visual'
+        atom.keymaps.cancelPendingState()
+        atom.beep()
 
     @editorElement.classList.add("vim-mode")
     @setupCommandMode()
--- a/lib/vim-state.coffee
+++ b/lib/vim-state.coffee
@@ -38,8 +38,8 @@ class VimState
 
     @subscriptions.add atom.keymaps.onDidFailToMatchBinding (e) =>
       return unless e.keyboardEventTarget is @editorElement
+      atom.keymaps.cancelPendingState()
       if @mode is 'operator-pending'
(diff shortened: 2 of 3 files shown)


bronson:
I had a problem with this PR in vim-mode-next.  `dd` didn't work anymore (just ignored).  Might be an issue with the way I merged it?

Here's the bad tree: https://github.com/bronson/vim-mode-next/commits/no-dd

And here's the revert that fixes it: https://github.com/bronson/vim-mode-next/commit/6105cea8ff7c4f9dcee14b3258ff85acde17fb8d

jacekkopecky:
@bronson I can't replicate it. Are specs failing, or is it interactive behaviour? Do you have anything with `d` in your own keymap, by any chance?

bronson:
No specs failing, I emptied my keymap.  I'll try disabling extensions...

Yep!  It only happens when vim-surround is installed.  That makes sense.  I should have tried disabling extensions before reporting.  Sorry, getting lazy.  :)

bronson:
Hm, vim-surround is starting to look notorious: https://github.com/atom/vim-mode/issues/657#issuecomment-122571442

I wish I could mark packages as conflicting...  I don't want to break people using both vim-mode-next and vim-surround, but I don't want vim-surround's nasty keymap handling to hold up progress either...  Dunno.

jacekkopecky:
Oh boy, this complicates things. Vim-surround has a key binding for `d s`, so vim-mode's `d d` becomes a binding atom fails to match. Let's see if vim-surround's maintainers are receptive to rethinking their key binding approach.

jacekkopecky:
No, I can't seem to formulate this as an issue against vim-surround - it's making a valid use of atom's key binding infrastructure, after all. The above commit addresses these particular interactions, but I'm no longer sure that this approach to handling unrecognised commands is appropriate in Atom vim-mode.

How about alternatively I work on showing the commands awaiting completion and if a user types `vix` or `ci3w` and gets bitten, they have `u` or `b` to help, and they'll learn that's not a valid combination. 

Maybe a debounced beep on unrecognized single-key bindings would help. I'll look into that too.

bronson:
Yeah, too bad this can't just be pinned on vim-surround.  Would have made things easy!  :)

You don't think it makes sense to just rework vim-surround's mappings to allow this, even though what they're doing now is OK?  If it makes things work and keeps the code clean then that seems reasonable...

jacekkopecky:
Thinking about your comments @bronson made me realize that perhaps #764 should only cancel on single-key unrecognized key bindings. 386e0c8 implements that.

Now this PR implements the following behaviour:
 * `ci3w` behaves as `3w` after canceling `c` because of unrecognized `i` in operator-pending mode (replayed by Atom after matching unrecognized `i3`). `ci3` beeps and means `3`.
 * `cqw` behaves as `w` after canceling `c` because of unrecognized `q`. `cq` beeps and is a no-op.
 * `vix` behaves as `vx`, but it beeps on the unrecognized `ix` (because atom replays `i` which goes unrecognized, and then it replays `x`).

I also plan to expand the specs to spy on beeping, so `apm test` doesn't beep and the specs verify that the beeps happen.

Thoughts welcome.

jacekkopecky:
@bronson with the current #764, the interaction with vim-surround looks like this: 
 1. you press `d`, Atom sees the possible `ds` from vim-surround so waits for a bit in case you press something. Vim-mode doesn't know yet that anything is happening.
 2. if you wait, Atom will time out waiting for `ds` and just play `d`, which puts us in the operator-pending mode. Vim-surround's complex key bindings cause a delay before the user sees operator-pending mode (and the associated cursor change).
 3. pressing `d` quickly after the first `d`, before Atom times out as described before, means Atom fails to find a multi-keystroke binding that starts with `dd` so it replays `d`, that's recognized by vim-mode which goes operator-pending, and then Atom replays the second `d` immediately and vim-mode deletes the line.

In short, waiting for complex key strokes delays user feedback, and that can be an issue for vim-surround. I'm tired now but will open it tomorrow if you don't do it first. 8-)

lee-dohm:

PR #765, opened 2015-07-12
repeating without search history should do nothing

jacekkopecky:
When Atom starts, there is no search history, and `n` would advance by a character because it searches for nothing, which matches everywhere.
This PR fixes that.


jacekkopecky:
another low-hanging, clean-merging fruit here

Issue #741, opened 2015-07-06
Throws error when deleting 2 paragraphs `d2ap`


zaz:
### Steps to reproduce:
1. Open new file
2. Create 2 paragraphs
3. <kbd>ggd2ap</kbd> (go to beginning and attempt to delete 2 paragraphs)

**Expected behaviour**: The two paragraphs disappear.
**Actual behaviour**: Throws error.

The same problem occurs with <kbd>d2ip</kbd>, and <kbd>2dap</kbd> does not delete both paragraphs.

**Atom Version**: 0.200.0
**System**: linux 3.18.1-1-chromebook
**Thrown From**: [vim-mode](https://github.com/atom/vim-mode) package, v0.54.0
### Stack Trace

Uncaught TypeError: undefined is not a function

```
At /home/zaz/.atom/packages/vim-mode/lib/vim-state.coffee:263

TypeError: undefined is not a function
  at Repeat.Prefix.isLinewise (/home/zaz/.atom/packages/vim-mode/lib/prefixes.coffee:30:21)
  at Delete.Operator.setTextRegister (/home/zaz/.atom/packages/vim-mode/lib/operators/general-operators.coffee:51:17)
  at Delete.execute (/home/zaz/.atom/packages/vim-mode/lib/operators/general-operators.coffee:96:8)
  at VimState.module.exports.VimState.processOpStack (/home/zaz/.atom/packages/vim-mode/lib/vim-state.coffee:266:23)
  at VimState.module.exports.VimState.processOpStack (/home/zaz/.atom/packages/vim-mode/lib/vim-state.coffee:258:10)
  at VimState.module.exports.VimState.processOpStack (/home/zaz/.atom/packages/vim-mode/lib/vim-state.coffee:258:10)
  at VimState.module.exports.VimState.pushOperations (/home/zaz/.atom/packages/vim-mode/lib/vim-state.coffee:224:8)
  at atom-text-editor.commands.(anonymous function) (/home/zaz/.atom/packages/vim-mode/lib/vim-state.coffee:196:45)
  at CommandRegistry.module.exports.CommandRegistry.handleCommandEvent (/usr/share/atom/resources/app.asar/src/command-registry.js:238:29)
  at /usr/share/atom/resources/app.asar/src/command-registry.js:3:61
  at KeymapManager.module.exports.KeymapManager.dispatchCommandEvent (/usr/share/atom/resources/app.asar/node_modules/atom-keymap/lib/keymap-manager.js:519:16)
[...]

PR #717, opened 2015-06-25
Whole word text objects aW/iW

jacekkopecky:
This PR also documents that we have text objects and which ones those are.

PR #623, opened 2015-04-12
Fix Till when selecting

jacekkopecky:
Till is an exceptional movement in that when selecting, it selects the current character even though it doesn't move the cursor. This was failing the normal selection machinery that expects cursor movements.
Fixes #618.

PR #766, opened 2015-07-12
fix: isLinewise shouldn't be required

### LINKED ISSUE #760: Crash on complex command
I tried to do the an error report/copy to clipboard, but only the toggle button did anything. Restarting atom didn't help. Screenshot:

![error][1]

Steps:

- <kbd>Cmd</kbd><kbd>N</kbd>
- <kbd>c</kbd><kbd>3</kbd><kbd>i</kbd><kbd>w</kbd>

I meant `ci3w` (which works fine). Version 0.54.0.

<sub>By the way, thanks for the best vim bindings out of any editor I've used!</sub>

[1]: http://i.imgur.com/S51zZhx.png

jacekkopecky:
This PR addresses the issue that generated an exception in #760 and #741. It doesn't add counts to text objects, that can be another PR; but at least vim-mode doesn't crash anymore on `c3aw`.
We might close #760 and #741 and open an issue that text objects don't have counts.

### DIFF
--- a/lib/prefixes.coffee
+++ b/lib/prefixes.coffee
@@ -27,7 +27,7 @@ class Prefix
     @composedObject.select?(@count)
 
   isLinewise: ->
-    @composedObject.isLinewise()
+    @composedObject.isLinewise?()
 
 #
 # Used to track the number of times either a motion or operator should


jacekkopecky:
This is a very low-hanging fruit (one line changed), merges cleanly with #573, #623, #711 and #717.

Issue #377, opened 2014-08-12
Camelcase movement operators


jspahrsummers:
Given the programming languages I typically work in, one of my favorite Vim plugins is [camelcasemotion](http://www.vim.org/scripts/script.php?script_id=1905). I'd love to build something similar for Atom, but it seems like it would depend on support within vim-mode directly.

Is there any precedent for implementing non-standard movements like this, or does anyone have recommended reading for building a separate package that relies upon vim-mode?

/cc @nathansobo @ymendel @bhuga @mcolyer

dsandstrom:
I'll start adding support for this.  However, what should the keymaps be?  I was using `q`, which works great, but that would conflict with macros if they are ever added.  I would like to use a single key, but there are no keys available that I know about.

deiwin:
The vim plugin seems to sacrifice the comma. I don't know if that's the best idea, though.

deiwin:
The `\` and `|` characters seem to be available.

jacekkopecky:
Just a thought - I could imagine a configuration option, or a per-language option somehow, for `w` and `b` to mind camelCase. Additionally, this setting could somehow affect the `iw` and `aw` text objects. 

It's about time (and this might be the place) that someone reinvents the principles of modal text editing in the modern setting of windows, panes, big screens, widespread camelCase, colours, highlights and so on.

deiwin:
This would probably mean that support for `iW` and `aW` should also be added to differentiate between the 2 possible meanings of *a word*. But I would probably like the uppercase versions to behave as the current lowercase versions do, instead of making them work as they currently do in *vim* (using the WORD definition).

Or maybe just have different bindings for all 3 cases. I mean, we have more modifier keys available than just shift.

lee-dohm:
As stated in [the README](https://github.com/atom/vim-mode), this package is no longer maintained and is deprecated. We recommend that people use the [vim-mode-plus package](https://atom.io/packages/vim-mode-plus) instead. Because of this, we are [archiving this repository](https://help.github.com/articles/about-archiving-repositories/) and closing all issues and pull requests. Thanks very much for your support and contributions!

PR #769, opened 2015-07-12
Add subword movement support

deiwin:
I will be fleshing out some ideas for #377 in this PR. Comments and feedback are welcome. When all of the main changes are made and you agree and are happy with them, I can clean up the commit history and add the necessary documentation etc.

The general idea is that there will be 3 different definitions of a word. The main one, an alternative one and the space delimited (*WORD*) one. The only difference between the main and the alt one would be that one of them would be sensitive to camel (and snake?) casing and the other would not.

TODO:
 - [x] add a configuration option that says whether a CamelCase sensitive or insensitive definition of a *word* is used by default
 - [ ] add an alternative version for the <kbd>w</kbd> movement
   - [ ] revert or re-fix https://github.com/atom/atom/commit/ba3ab41
   - [ ] fix the tests that fail when <kbd>Alt</kbd> is used as a modifier
   - [ ] wait until atom/atom#7658 has been fixed and see if it fixes the already added tests
 - [ ] add an alternative version for the <kbd>b</kbd> movement
 - [ ] add an alternative version for the <kbd>e</kbd> movement
 - [ ] add an alternative version for the <kbd>w</kbd> text object
 - [ ] **?** add an alternative version for the <kbd>Ctrl</kbd>-<kbd>w</kbd> binding in the insert mode

### DIFF
--- a/keymaps/vim-mode.cson
+++ b/keymaps/vim-mode.cson
@@ -90,6 +90,7 @@
   'e': 'vim-mode:move-to-end-of-word'
   'E': 'vim-mode:move-to-end-of-whole-word'
   'b': 'vim-mode:move-to-previous-word'
+  'alt-b': 'vim-mode:move-to-previous-alt-word'
   'B': 'vim-mode:move-to-previous-whole-word'
   '}': 'vim-mode:move-to-next-paragraph'
   '{': 'vim-mode:move-to-previous-paragraph'
--- a/keymaps/vim-mode.cson
+++ b/keymaps/vim-mode.cson
@@ -85,6 +85,7 @@
   'down': 'vim-mode:move-down'
 
   'w': 'vim-mode:move-to-next-word'
+  'alt-w': 'vim-mode:move-to-next-alt-word'
   'W': 'vim-mode:move-to-next-whole-word'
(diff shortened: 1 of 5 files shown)


deiwin:
atom/atom@ba3ab41 from atom/atom#7612 makes the subword regex match line endings if there's any trailing whitespace and line beginnings if it's followed by any whitespace. I currently worked around it, but I think the regex should be changed.

deiwin:
Can someone help me figure out why some of the tests fail when I use alt as a modifier. They seem to work just fine when actually used. I tried messing around with the way the spec-helper dispatches keyboard events, but to no avail.

jacekkopecky:
This is a guess: it seems to me that CI is complaining about the presence of focused specs. Please try changing `fdescribe` to `describe`.

jacekkopecky:
I like the procedurally generated tests with `itMovesByWord` and `itMovesBySubWord`, by the way!

deiwin:
I'd kept the focused specs because I hadn't really gotten them to pass on my own machine yet and wasn't worrying too much about the CI build. I added a commit to change that now, however, so that the CI logs would show the actual problem I'm facing.

deiwin:
Not working on this anymore

Issue #579, opened 2015-03-14
mouse clicks position command-mode cursor after end of line


jacekkopecky:
Clicking the mouse to position the cursor can easily place it after the last character, while it should put it **on** the last character as all command-mode motions do.

As VIM has its command-mode cursor **on** a character, rather than between characters, it doesn't normally allow the cursor to be after the last character.

hailiang:
+1

Issue #2, opened 2013-08-01
Prevent selection of line-returns in command mode


mcolyer:
In vim, you can only select line returns when in insert mode, as specified in this spec https://github.com/atom/vim-mode/blob/v0.1.0/spec/vim-mode-spec.coffee#L50-L59

I'm not exactly sure the best way to handle this given the current editor API. @probablycorey or @nathansobo any ideas on how to approach this?

bronson:
Does this issue need to be fixed?  It seems like vim-mode is currently doing a great job of not allowing the cursor to ever reach the newline.  Do we need another layer of defense?

bronson:
aha, clicking can place the cursor over the newline.  Very few vimmers will notice.  :)  Anything else?

jacekkopecky:
I have noticed and it's annoying, it's on my list but nowhere near the top.

PR #770, opened 2015-07-12
use onDidChangeCursorPosition for keeping cursor in line

jacekkopecky:
Currently, motions would individually ensure that cursor doesn't end up after the last character of a line. This fails in a number of circumstances, e.g. with mouse clicks to move/add a cursor, and with specs that `setCursorBufferPosition` to put the cursor where it should never be.

This PR reacts to any cursor movement with TextEditor's `onDidChangeCursorPosition` and any cursor addition with `onDidAddCursor` and makes sure the cursor never goes where it doesn't belong.

The PR also fixes some tests that expected the cursor after the last character in command mode, and removes some outdated spec comments.

### DIFF
--- a/lib/vim-state.coffee
+++ b/lib/vim-state.coffee
@@ -39,7 +39,7 @@ class VimState
         @activateVisualMode('characterwise') if @mode is 'normal'
     , 100)
 
-    @subscriptions.add @editor.onDidChangeCursorPosition @ensureCursorIsWithinLine
+    @subscriptions.add @editor.onDidChangeCursorPosition ({cursor}) => @ensureCursorIsWithinLine(cursor)
     @subscriptions.add @editor.onDidAddCursor @ensureCursorIsWithinLine
 
     @editorElement.classList.add("vim-mode")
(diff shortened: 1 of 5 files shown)


jacekkopecky:
This PR also fixes #2, as far as I can tell. Yay for fixing the longest-outstanding issue of vim-mode. 8-)

bronson:
This is terrific!  It's nice when the fix also deletes code scattered all over the place.

Minor conflicts with https://github.com/atom/vim-mode/pull/591.  Both should be merged.

jacekkopecky:
fixes #579

lib/vim-state.coffee
```diff
@@ -654,6 +662,16 @@ class VimState
     text = @getRegister(name)?.text
     @editor.insertText(text) if text?
 
+  ensureCursorIsWithinLine: (cursor) =>
+    return if @processing or @mode is 'visual' or @mode is 'insert'
```
maxbrunsfeld: Can you explain a bit about why this `@processing` variable? Without looking into this too deeply, I would think that we could just `return unless @mode is 'normal'`.

lib/vim-state.coffee
```diff
@@ -654,6 +662,16 @@ class VimState
...
+
+    if cursor.cursor? then cursor = cursor.cursor # when called by @editor.onDidChangeCursorPosition
```
maxbrunsfeld: I think this would be a little more straightforward if you removed this line, and when observing the cursor position did this:

```coffee
@editor.onDidChangeCursorPosition ({cursor}) => @ensureCursorIsWithinLine(cursor)
```

maxbrunsfeld:
Very good idea.

lib/vim-state.coffee
jacekkopecky: this is nice!

lib/vim-state.coffee
jacekkopecky: Some motions, esp. with counts, invoke several Atom motions, each notifies us about cursor movement, and they all can happen in normal mode. We don't want to interfere until that's done. 

But since it needs explanation, it could be written clearer or better; suggestions welcome.

jacekkopecky: But 'not "normal"' should be OK instead of two comparisons

jacekkopecky: that's in the new commit, which also adds *more* `@processing` 8-)

maxbrunsfeld: > We don't want to interfere until that's done.

I see. I think this is a fine solution actually.

maxbrunsfeld:
Ok great! Congratulations on closing issue #2.

PR #772, opened 2015-07-15
clean up: remove unused selectOptions and allowEOL

jacekkopecky:
We never do use those parameters. Adapting #581 to not use selectOptions is quite simple.


maxbrunsfeld:
Thanks for :fire:ing those.

PR #774, opened 2015-07-15
Canceling c operator

jacekkopecky:
When a motion matches nothing, e.g. `%` on a line without brackets, or `fx` on a line without x, `c` should be cancelled and go into command mode; currently it simply behaves like `i`. 
This PR implements the desired behaviour.

Also, this PR adds a lot of tests for `c%`; I wrote those when I mistakenly thought there was some other problem there so I wanted to document the desired behaviour and fix it, and then I found it already worked as I'd described it. If the tests are undesirable, I can remove them.

### DIFF
--- a/lib/operators/input.coffee
+++ b/lib/operators/input.coffee
@@ -111,11 +111,11 @@ class Change extends Insert
   #
   # Returns nothing.
   execute: (count) ->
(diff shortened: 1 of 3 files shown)


jacekkopecky:
will fix specs after #787 is merged

maxbrunsfeld:
Great!

PR #776, opened 2015-07-17
hide input outline

t9md:
I use one-dark-ui or one-light-ui theme.
For  mini editor used in `f` command, outline is shown bottom of window.

This PR hide this outline.

image

### DIFF
--- a/styles/vim-mode.less
+++ b/styles/vim-mode.less
@@ -11,6 +11,9 @@
   cursor: default;
   white-space: nowrap;
   padding-left: 10px;
+  &.is-focused {
+    box-shadow: none;
+  }
 }
 
 .block-cursor(@visibility: visible) {


jacekkopecky:
This also removes the blue box from the search box, though. Also, what does it look like on your system when you press `m` for *create mark*? I've created an alternative, much more complicated take on this issue: #777

t9md:
When pressing `m` , input area is still shown on bottom area( not completely hidden ).
I'm happy if complete solution is there, my PR is simply modifying style its can be done on user's style.less.

But merging this PR is no harm since its original intent is hide margin and outline.

jacekkopecky:
@t9md I see your point. My theme doesn't use such outline so it doesn't affect me, but I'd probably see such visual tweaks as the domain of themes rather than packages. I'd expect the blue outline to be useful for the search field, for consistency with the rest of the UI. But again, my theme doesn't use box-shadow there.

t9md:
> I'd probably see such visual tweaks as the domain of themes rather than packages

I agree, but for vim-mode specific field its kind of domain vim-mode allow to modify it.
If search field want to respect UI theme is should use  separate editor form than f command at first.

Anyway vim-mode's these helper input field need to be improved further than this quick PR.
I want your PR merged quickly.

Thanks for a lot of improving to vim-mode always.

jacekkopecky:
Hi, since #777 was merged, should this one be closed?

PR #778, opened 2015-07-17
add operation yank-to-last-character-of-line

diogocampos:
...to allow mapping `Y` to do this instead of yanking the entire line.

### DIFF
--- a/lib/vim-state.coffee
+++ b/lib/vim-state.coffee
@@ -92,6 +92,7 @@ class VimState
       'toggle-case-now': => new Operators.ToggleCase(@editor, this, complete: true)
       'yank': => @linewiseAliasedOperator(Operators.Yank)
       'yank-line': => [new Operators.Yank(@editor, this), new Motions.MoveToRelativeLine(@editor, this)]
+      'yank-to-last-character-of-line': => [new Operators.Yank(@editor, this), new Motions.MoveToLastCharacterOfLine(@editor, this)]
       'put-before': => new Operators.Put(@editor, this, location: 'before')
       'put-after': => new Operators.Put(@editor, this, location: 'after')
       'join': => new Operators.Join(@editor, this)


mattr-:
Any reason you can't create a mapping to do this with the existing yank command and a motion like you would in vim? (`map Y y$`)

jacekkopecky:
I suspect the reason is that composing vim-mode's commands can really only be done inside vim-mode; perhaps we need some kind of API for that? Time to look at what those *services* are all about.

jacekkopecky:
Oh, combining motions outside of vim-mode is possible: here's what I have in my `init.coffee` to combine `n` and `zz`:
```
atom.commands.add 'atom-text-editor.vim-mode.command-mode', 'jacek:center-next', ->
  editor = atom.workspace.getActiveTextEditor()
  editorElement = atom.views.getView(editor)
  vimState = VimMode.getEditorState(editor)
  vimState.pushOperations new Motions.RepeatSearch(editor, vimState)
  vimState.pushOperations new Scroll.ScrollCursorToMiddle(editorElement, {leaveCursor: true})
```
And then in the `keymap.cson`:
```
'atom-text-editor.vim-mode:not(.insert-mode)':
  '\\': 'jacek:center-next'
```
This could, however, be easier.

jacekkopecky:
I think this should be possible in the author’s own configuration by combining `y` and `$`. Currently there's a way outlined above; but should vim-mode provide an API akin to vim’s `:map`?

mattr-:
Vim doesn't provide one, so my vote (initially) is 'no'

mattr-:
Oh, wait, I misunderstood the question earlier. Initially, I said 'no' because I wasn't in agreement with providing a default mapping for `Y`. That being said, I'd love to see a `:map` command implemented in vim-mode, but Atom's already existing facilities are useful for this too, so ¯\_(ツ)_/¯

romgrk:
@mattr- I do have a personnal `:map` command but it's not production level.
It allows to write mappings like:
```coffeescript
maps =
    '<a-k>': '5k'
    '<a-j>': '5j'
    '<a-d>': '<c-d>'
    '<a-u>': '<c-u>'

nmaps =
    '<CR>':   'o<esc>'
    '<S-CR>': 'O<esc>'
    'Y':    'y$'
    'S':    'v$hs'
```
If you're interested I could try to bundle something.

lee-dohm:

PR #777, opened 2015-07-17
Hide command input field better

jacekkopecky:
Also removed the seemingly unnecessary editorContainer; this may
affect people’s personal styles if anybody has applied styles to that
container element.

The change in spec-helper makes sure that the new code’s expectations
are met when testing.

This is an alternative solution to PR #776.

PR #779, opened 2015-07-18
give operator-pending mode a rectangle cursor

bronson:
As discussed in https://github.com/atom/vim-mode/pull/777

### DIFF
--- a/styles/vim-mode.less
+++ b/styles/vim-mode.less
@@ -28,6 +28,14 @@
   opacity: 0.5;
 }
 
+.low-block-cursor(@visibility: visible) {
+  border: none;
+  border-bottom: 10px solid @syntax-cursor-color;
+  background: none;
+  visibility: @visibility;
+  opacity: 0.5;
+}
+
 atom-text-editor.vim-mode.normal-mode,
 atom-text-editor.vim-mode.operator-pending-mode,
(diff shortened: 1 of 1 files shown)


bronson:
I like the hollow rectangle but Jack's original version has a thick-underline cursor:

```less
.half-cursor(@visibility: visible) {
  border: none;
  border-bottom: 10px solid @syntax-cursor-color;
  background: none;
  visibility: @visibility;
  opacity: 0.5;
}
```

Happy to go with whatever people think looks good.

bug3 bug3

bronson:
Now we just need an expecting-a-character cursor that can be used for `f`, `t`, and register commands.

styles/vim-mode.less
```diff
@@ -20,6 +20,15 @@
   opacity: 0.5;
 }
 
+.low-block-cursor(@visibility: visible) {
+  border: none;
+  border-bottom: 10px solid @syntax-cursor-color;
```
jacekkopecky: may I suggest .6em rather than 10px here? When I pasted my custom styles, I didn't remember to change to relative units.

bronson:
Agreed, updated.

maxbrunsfeld:
:clap: Nice.

bronson:
Rebased to master.

PR #780, opened 2015-07-18
fix `cc`, make `S` an alias for `cc`

### LINKED ISSUE #625: `cc` should not require me to reset the indentation
I think this may be a bug on Atom core because vim-mode seems to simply call out to it.

jacekkopecky:
Fix #625. Makes `S` an alias for `cc` and fixes `cc` to do what `S` used to do, when combined with linewise motions.


maletor:
:100: Thanks!

PR #781, opened 2015-07-18
beep on unrecognized commands to warn the user

jacekkopecky:
Alternative to #764. Instead of trying to cancel an operation, possibly running into conflict with other packages' key bindings (as discussed in #764), this PR just beeps on unrecognised key bindings. 

Together with some form of showing the current pending key strokes (thought in progress), this could be a way of handling unrecognised commands that's more appropriate for Atom, even though it doesn't behave exactly like VIM.

Discussion welcome.

### DIFF
--- a/lib/utils.coffee
+++ b/lib/utils.coffee
@@ -1,5 +1,8 @@
 {Range} = require 'atom'
 
+# copied from atom/atom-keymap src/helpers.coffee
+AtomModifierRegex = /(ctrl|alt|shift|cmd)$/
+
 module.exports =
   # Public: Determines if a string should be considered linewise or character
   #
@@ -25,3 +28,8 @@ module.exports =
       newRange
     else
       oldRange.union(newRange)
+
+  # copied and simplified from atom/atom-keymap src/helpers.coffee
+  # see atom/atom-keymap#97
+  isAtomModifier: (keystroke) ->
+    AtomModifierRegex.test(keystroke)
--- a/lib/vim-state.coffee
+++ b/lib/vim-state.coffee
@@ -44,6 +44,12 @@ class VimState
     else
       @activateNormalMode()
 
+    @subscriptions.add atom.keymaps.onDidFailToMatchBinding (e) =>
+      return unless e.keyboardEventTarget is @editorElement
+      return if Utils.isAtomModifier(e.keystrokes)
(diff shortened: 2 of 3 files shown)


bronson:
Let me see if I understand...  The problem with #764 was that vim-mode thinks that it knows all possible key combinations...  So, if vim-surround adds a new combination, vim-mode wouldn't recognize it and cancel it before it's complete?

Wouldn't this fix suffer the same issue...?  It would beep every time it sees the unrecognized combo, even though vim-surround would do the right thing?

jacekkopecky:
This one only beeps when a single-key keystroke hasn't been matched. The problem with #764 is that it needs to react to multi-key keystrokes. Examples:

In `ci3w`, it sees `i3` as a key binding not recognized by Atom; and Atom only attempts to match `i3` because other two-key strokes starting with `i` exist. #764 cancels `i3` and cancels `c` as well, in order to prevent replaying `i` (not recognized so no-op) and `3` which then gets tied to `w`.

With vim-surround, the existence of the key strokes starting with `ds` mean that when the user types `dd`, Atom waits after the first `d` in case `s` comes, and when it doesn't, Atom shouts that `dd` was not recognized, #764 cancels the whole thing and the two `d`s don't get replied, to be combined by vim-mode.

In here - #781 - there's no canceling, just beeping, and only on single-key unmatched bindings.

Which brings me to an idea that #764 could only cancel on single-key unmatched bindings. That should take care of interactions with vim-surround. We'll see momentarily.

jacekkopecky:
rebased and ready for your attention @maxbrunsfeld (after #808, #827, and #783 please)

I think I prefer this approach (beep on unrecognized key strokes, without canceling any) to the approach of #764 which tries to cancel pending commands but may delay user feedback.

Neither of these two PRs can copy VIM's behaviour, but it's about handling erroneous input so there should be no consideration of muscle memory.

jacekkopecky:
oh boy, tiny merge conflict with #827 - I'll be happy to rebase as soon as that one is merged

jacekkopecky:
OK, this one is now ready; please see my [comment above](https://github.com/atom/vim-mode/pull/781#issuecomment-133180031) when considering this.

lee-dohm:

PR #783, opened 2015-07-19
Inclusiveness fixes

### LINKED ISSUE #753: % motion composing with d command
I've been working on a rebuild of the % bracket matching system to support additional types for comments and functions. I've come across a bug that seems to work oddly with the `operatesInclusively` bit. Before I try to dig in, I wanted to see if anyone had any thoughts!

In a blank file, input `(test)` and move the command-mode cursor to the closing parentheses. Use the command `d %` to delete to the matching open parentheses. You should be left with an empty file, but instead the closing item remains!

Any thoughts?

jacekkopecky:
This PR addresses one functionality issue with inclusive motion backwards: inclusive movement backwards should include the character under cursor. 

At the moment in vim-mode, only `%` is an inclusive movement that can go backwards, so only it is affected. When search gets postfixes (basically `/e`), or when we implement operator-pending `v` motions (e.g. `dvb` is an inclusive version of `db`), other motions too will be affected.

The PR is seemingly more involved than it would have to be, but IMHO it includes useful cleanups:
 1. the `operatesInclusively` flag in `Motion` should be `false` by default because inclusive motions are the exception, the norm is exclusive motions.
 2. I removed the `operatesInclusively` flag where it's redundant.
 3. I distinguished between non-visual selection for inclusive movements (the rewritten `moveSelectionInclusively` method on `Motion`) and selection movement in visual mode (the new `moveSelectionVisual`, which is the old `moveSelectionInclusively`): the old logic was really for any movement in visual mode, and while one could amend it to handle the backward exclusive non-visual movements correctly, I really saw it as combining two functionalities in one unnecessarily.
 3. I understood and added comments in the old logic (now `moveSelectionVisual`) - it was pretty dense code.

The specs are in common with #774, just tweaked for what's fixed here.

Fixes #753.

### DIFF
--- a/spec/operators-spec.coffee
+++ b/spec/operators-spec.coffee
@@ -737,8 +737,7 @@ describe "Operators", ->
           keydown('c')
           keydown('%')
           editor.insertText('x')
-          # this differs from VIM, which deletes the character originally under cursor
-          expect(editor.getText()).toBe("12345x67)8\nabcx)e\nAx)BCDE")
+          expect(editor.getText()).toBe("12345x7)8\nabcxe\nAxBCDE")
           expect(vimState.mode).toBe "insert"
 
       describe "after or without brackets", ->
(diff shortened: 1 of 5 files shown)


spec/operators-spec.coffee
```diff
@@ -680,6 +680,88 @@ describe "Operators", ->
...
+          editor.setText("12345(67)8\nabc(d)e\nABCDE")
+          editor.setCursorScreenPosition([0, 9])
+          editor.addCursorAtScreenPosition([2, 2])
+          keydown('c')
+          keydown('%')
+          expect(editor.getText()).toBe("12345(67)8\nabc(d)e\nABCDE")
+          # differs from VIM, fixed in #774
+          expect(vimState.mode).toBe "insert"
+
+      describe "repeats correctly with .", ->
```
maxbrunsfeld: Maybe this should say something general, like `describe "repetition and undo", ->`.

jacekkopecky: quite right, sorry, don't know what happened there; restructured now.

jacekkopecky:
Ah, specs tripped by the command/normal rename; I'll fix that after #787

jacekkopecky:
all new and rebased now

maxbrunsfeld:
Great!

PR #784, opened 2015-07-20
Protect against edge cases with pane splits

### LINKED ISSUE #657: Operators replace text with motion when used with pane splits

jazzpi:
When splitting a pane, input would be enabled in the original pane in command
mode (Fixes #657)

Not really sure if the spec is in the right place :confused:

### DIFF
--- a/lib/vim-mode.coffee
+++ b/lib/vim-mode.coffee
@@ -31,6 +31,19 @@ module.exports =
     @disposables.add new Disposable =>
       @vimStates.forEach (vimState) -> vimState.destroy()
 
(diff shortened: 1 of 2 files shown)


bronson:
When I comment out the new code in vim-mode.coffee, the spec still passes.

Does it fail on your machine?

lee-dohm:

Issue #598, opened 2015-03-30
Change/delete inner tag not working correctly


isaachess:
Extension of issue #418.

Change/delete inner tag is not working like it does in vim. It should detect the tag it is in (or the _nearest_ tag, as vim does), then change/delete all content within that tag.

Currently it appears to delete everything until the next tag it finds, even if that tag isn't the same as the parent. Really buggy right now.

lee-dohm:
As stated in [the README](https://github.com/atom/vim-mode), this package is no longer maintained and is deprecated. We recommend that people use the [vim-mode-plus package](https://atom.io/packages/vim-mode-plus) instead. Because of this, we are [archiving this repository](https://help.github.com/articles/about-archiving-repositories/) and closing all issues and pull requests. Thanks very much for your support and contributions!

PR #795, opened 2015-07-27
Add SelectInsideTags and SelectAroundTags text-objects

MarkusSN:
This adds support for the tags text objects, but currently only works for single line tags. I am planning on adding support for multiline tags, but I decided some early feedback would be useful.


MarkusSN:
Added support for multiline tags.

MarkusSN:
#598

spec/text-objects-spec.coffee
```diff
@@ -164,28 +164,232 @@ describe "TextObjects", ->
...
+
     beforeEach ->
-      editor.setText("<something>here</something><again>")
-      editor.setCursorScreenPosition([0, 5])
+      editor.setText('clean')
+      editor.setCursorBufferPosition([0, 0])
+      keydown('y')
+      keydown('i')
+      keydown('w')
+      expect(vimState.getRegister('"').text).toBe('clean')
```
bronson: I don't understand this...?  Looks like you're testing yanking inside a word before each inner tag test?

MarkusSN:
Lol, I understand the confusion. I just wanted to make sure that I reset the register before each test, and that was the first thing I came up with. The plan was to find a better way to set the value in the register, but I obviously forgot :)

spec/text-objects-spec.coffee
MarkusSN: Lol, I understand the confusion. I just wanted to make sure that I reset the register before each test, and that was the first thing I came up with. The plan was to find a better way to set the value in the register, but I obviously forgot :)

MarkusSN:
Currently, the implementation assumes a correctly formatted xml structure. But it seems like unclosed tags are ignored in vim, so I am going to try to get that working.

MarkusSN:
Rebased.
Seems to working pretty well now. It's a bit massive compared to some of the other text objects, so if anyone has any feedback, that would be great :)

jacekkopecky:
Would this address #835?

MarkusSN:
Yes :)

MarkusSN:
I changed one of the existing tests to expect that "it does _not expand_ an existing selection in visual mode" instead of "it expands an existing selection in visual mode". This is more in line with how it works in vim.

lee-dohm:

Issue #796, opened 2015-07-27
Enable/Disable vim-mode based on file type? (e.g. markdown)


chibicode:
Hi, I love vim-mode but was wondering if there's an easy way to enable/disable vim-mode for certain file types, like markdown. If not, a setting entry for ignored file names like the screenshot below would be really awesome.

![image](https://cloud.githubusercontent.com/assets/992008/8916460/98cc4b74-3460-11e5-9f96-dc351ae97e26.png)

Sorry if this is a duplicate issue - my search for "file type enable" and "file type disable" didn't yield relevant results.

Just FYI, I'd like to have it disabled on markdown files b/c lines can get really long in markdown and need soft wrap, which isn't supported by vim-mode, according to README: https://github.com/atom/vim-mode/commit/eb0864f06465af4f764d14cfe7a7daed8fa60eba

maxbrunsfeld:
Vim-mode should support soft-wrap. We should remove that paragraph from the README. Have you found any cases where soft-wrapped lines are still treated incorrectly?

Regarding your initial question, I don't think there is a way to do this. However, you can deactivate and activate the package repeatedly, so it should be easy to write a script that deactivates vim-mode while a markdown editor is focused. See the `atom.workspace.observeActivePaneItem()` method.

chibicode:
@maxbrunsfeld thanks! Right, after some testing the only issue I came across w/ soft wrap was this: https://github.com/atom/vim-mode/issues/785

I made a README PR: https://github.com/atom/vim-mode/pull/797/files

> See the atom.workspace.observeActivePaneItem() method.

Thanks for the pointer! I'll look into this.

PR #797, opened 2015-07-27
Update README: Vim-mode should support soft-wrap

chibicode:
See: https://github.com/atom/vim-mode/issues/796#issuecomment-125340871


maxbrunsfeld:
Thanks!

t9md:
I believe this issue is only apparent in visual linewise interactive operation.
I described what is happening and fixed in my forked version of vim-mode.

Hope this information helps also for official vim-mode.

https://github.com/t9md/atom-vim-mode-plus/issues/8#issuecomment-144291851

PR #798, opened 2015-07-28
Indent operations preceded by a number should indent multiple times in visual mode.

mleeds95:
Preceding '>' with a number didn't indent lines multiple times, but this fixes that.

### DIFF
--- a/docs/operators.md
+++ b/docs/operators.md
@@ -26,7 +26,11 @@
 * Indent/Outdent/Auto-indent
   * `vw>` - works in visual mode
   * `>>` - indent current line one level
+  * `>2>` - repeated linewise
+  * `v2>` - repeat indent operation in visual mode
   * `<<` - outdent current line one level
+  * `<2<` - repeated linewise
+  * `v2<` - repeat outdent operation in visual mode
   * `==` - auto-indents current line
 * [Put](http://vimhelp.appspot.com/change.txt.html#p)
   * `p` - default register
--- a/lib/operators/indent-operators.coffee
+++ b/lib/operators/indent-operators.coffee
@@ -1,3 +1,4 @@
+_ = require 'underscore-plus'
 {Operator} = require './general-operators'
 
 class AdjustIndentation extends Operator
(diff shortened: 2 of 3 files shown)


maxbrunsfeld:
Could you add a small test for this, or modify an existing test so that it supplies a count?

mleeds95:
Sure. Do you have an automated way to run all the spec files?

maxbrunsfeld:
Yeah, you can run the `Window: Run Package Specs` from Atom, or run `apm test` from the command line.

docs/operators.md
```diff
@@ -26,7 +26,11 @@
 * Indent/Outdent/Auto-indent
   * `vw>` - works in visual mode
   * `>>` - indent current line one level
+  * `>2>` - repeated linewise
+  * `2>>` - repeat indent operation
   * `<<` - outdent current line one level
+  * `<2<` - repeated linewise
+  * `2<<` - repeat outdent operation
```
jacekkopecky: I tried it and in normal mode `>2>` behaves the same as `2>>` - indenting two lines once; that's consistent with VIM. The docs above seem to imply this PR was introducing a difference there - is it?

lib/operators/indent-operators.coffee
```diff
@@ -6,7 +7,10 @@ class AdjustIndentation extends Operator
     @motion.select(count)
     {start} = @editor.getSelectedBufferRange()
 
-    @indent()
+    if mode is 'visual'
+      _.times(count, => @indent())
```
jacekkopecky: needs `@editor.transact` before `_.times` - you may see that in `general-operators.coffee` - otherwise you need to press `u` twice to undo `v2>` .

mleeds95:
Thanks for the help @jacekkopecky. Should be fixed now.

bronson:
I had the same misconception as Jacek.  Just making sure, this PR:

- implements and documents v2> so it works the same as Vim
- it also documents 2>> and >2>, which currently work in vim-mode

Merged to vim-mode-next, looking good.

mleeds95:
Right. I should have made that more clear.

maxbrunsfeld:
:+1: Thanks @mleeds95! Nice work.

Issue #759, opened 2015-07-10
TypeError: undefined is not a function


codykrieger:
Atom: v1.0.1
vim-mode: v0.55.0

I was fiddling with a selection inside of a set of backticks in a ruby file when I saw this (something something (``vi` ``). I haven't been able to reproduce it with some (extremely) limited testing.

Atom originally pointed me at #705, which was already closed.

Backtrace:

```
TypeError: undefined is not a function
  at VimState.module.exports.VimState.processOpStack (/Users/cjk/.atom/packages/vim-mode/lib/vim-state.coffee:268:23)
  at VimState.module.exports.VimState.pushOperations (/Users/cjk/.atom/packages/vim-mode/lib/vim-state.coffee:226:8)
  at ViewModel.confirm (/Users/cjk/.atom/packages/vim-mode/lib/view-models/view-model.coffee:11:15)
  at vim-command-mode-input.VimCommandModeInputElement.confirm (/Users/cjk/.atom/packages/vim-mode/lib/view-models/vim-command-mode-input-element.coffee:46:16)
  at /Users/cjk/.atom/packages/vim-mode/lib/view-models/vim-command-mode-input-element.coffee:36:10
  at Emitter.module.exports.Emitter.emit (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/event-kit/lib/emitter.js:82:11)
  at TextBuffer.module.exports.TextBuffer.applyChange (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/text-buffer/lib/text-buffer.js:626:20)
  at TextBuffer.module.exports.TextBuffer.setTextInRange (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/text-buffer/lib/text-buffer.js:534:12)
  at Selection.module.exports.Selection.insertText (/Applications/Atom.app/Contents/Resources/app.asar/src/selection.js:468:43)
  at /Applications/Atom.app/Contents/Resources/app.asar/src/text-editor.js:736:29
  at /Applications/Atom.app/Contents/Resources/app.asar/src/text-editor.js:778:29
  at TextBuffer.module.exports.TextBuffer.transact (/Applications/Atom.app/Contents/Resources/app.asar/node_modules/text-buffer/lib/text-buffer.js:740:18)
  at TextEditor.module.exports.TextEditor.transact (/Applications/Atom.app/Contents/Resources/app.asar/src/text-editor.js:1151:26)
[...]

PR #805, opened 2015-07-31
Text Objects in normal mode without a preceding operator should select text.

mleeds95:
This should close issue #759. But I'm not entirely sure this code is the best way to deal with it so feedback is welcome!

### DIFF
--- a/lib/text-objects.coffee
+++ b/lib/text-objects.coffee
@@ -8,6 +8,8 @@ class TextObject
   isComplete: -> true
   isRecordable: -> false
 
+  execute: -> @select.apply(this, arguments)
+
 class SelectInsideWord extends TextObject
   select: ->
     @editor.selectWordsContainingCursors()
--- a/spec/text-objects-spec.coffee
+++ b/spec/text-objects-spec.coffee
@@ -21,6 +21,17 @@ describe "TextObjects", ->
   normalModeInputKeydown = (key, opts = {}) ->
     editor.normalModeInputView.editorElement.getModel().setText(key)
 
+  describe "Text Object commands in normal mode not preceded by an operator", ->
+    beforeEach ->
+      vimState.activateNormalMode()
+
+    it "selects the appropriate text", ->
+      editor.setText("<html> text </html>")
+      editor.setCursorScreenPosition([0, 7])
+      # Users could dispatch it via the command palette
+      atom.commands.dispatch(editorElement, "vim-mode:select-inside-tags")
+      expect(editor.getSelectedScreenRange()).toEqual [[0, 6], [0, 12]]
+
   describe "the 'iw' text object", ->
     beforeEach ->
       editor.setText("12345 abcde ABCDE")


maxbrunsfeld:
Thanks for taking the time to fix this! Could you add a spec in [text-objects-spec](https://github.com/atom/vim-mode/blob/master/spec/text-objects-spec.coffee) for this case?

jacekkopecky:
I could see an alternative solution to #759, if you don't mind:

Text Objects are all called `select-<something>` – they could be applicable in any mode to simply select whatever their selection is. This would mean TextObject to have the following single extra line:
```
  execute: -> @select.apply(this, arguments)
```

This way, Text Objects would be extra commands that vim-mode normal mode, and insert mode, doesn't have key bindings for, but they would still be meaningful.

mleeds95:
I agree, @jacekkopecky. I thought this project is trying to mimic vim's functionality, but if we're not bound to that I think that's a better solution.

maxbrunsfeld:
That sounds good.

@mleeds95 do you want to make this change on your branch?

mleeds95:
Yeah I'll rebase and force push.

mleeds95:
This makes the second `instanceof` check on [this line](https://github.com/atom/vim-mode/blob/master/lib/vim-state.coffee#L224) unnecessary. Should we take it out?

jacekkopecky:
I for one would leave it there.

mattr-:
I'm confused. It seems like this is fixing s bug in visual mode, but the description says that it allows text objects to select text in normal mode.

Honestly, it doesn't make much sense for text objects to select text in normal mode. For example, if I type `iw` (The inner word text object) in normal mode in Vim, I enter insert mode and type a 'w'. Does this PR actually change that behavior?

mleeds95:
@mattr- You're right that because the text object commands all start with i or a you would enter insert mode if you use the keyboard shortcuts. But if you execute the commands via the command palette instead (Ctrl + Shift + P) you stay in normal mode and vim-mode currently throws an error. This PR selects the text as if you're in visual mode, although it's debatable if we should just ignore the command.

mattr-:
I feel like we should just ignore the command, since it doesn't actually apply. Maybe the real fix is to set it up so that these are only applicable in operator-pending mode and therefore don't show up in the command palette for normal mode?

jacekkopecky:
@mattr- well, we can say the commands don't apply, or we can ask "could they be useful?" 

Currently, all vim-mode commands are shown in the command palette, even if they don't have key bindings in the current mode (e.g. the text objects in normal mode). Some might genuinely not apply, but I can't see an argument against this particular PR, making text objects select when invoked in normal mode.

mattr-:
If the selection only happens when they're invoked from the command palette, then I can live with that.

maxbrunsfeld:
Thanks @mleeds95; I don't feel strongly about the commands' behavior in normal mode, but it's great to stop this exception from happening to people.

PR #808, opened 2015-08-01
fix replace-mode backspace behavior to something acceptable

jacekkopecky:
Due to atom/text-buffer#81 and 30e4073c52f68221dccd3599dd60728f300b3534, backspace in replace mode hasn't been working well. 

For example, `ctrl-r x` to paste the `x` register would have backspace first undo the insert, then the delete, which looks wrong. Another example: when replacing after the end of the line, the grouping of multiple typed characters for undo confused our replace-mode backspace to no end.

### DIFF
--- a/lib/vim-state.coffee
+++ b/lib/vim-state.coffee
@@ -75,7 +75,7 @@ class VimState
       'repeat-prefix': (e) => @repeatPrefix(e)
       'reverse-selections': (e) => @reverseSelections(e)
       'undo': (e) => @undo(e)
-      'replace-mode-backspace': => @replaceModeUndo()
+      'replace-mode-backspace': => @replaceModeBackspace()
       'insert-mode-put': (e) => @insertRegister(@registerName(e))
       'copy-from-line-above': => InsertMode.copyCharacterFromAbove(@editor, this)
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Still thinking about this one. I think your solution with the transaction grouping interval is good, but I'd like to avoid adding more direct interrogation of the text-buffer's `undoStack`, because that code is private, so its implementation is subject to change. Maybe what this is telling us is that we need a new public method on `TextBuffer`, like `::hasChangesSinceCheckpoint(checkpoint)`. What do you think about that?

jacekkopecky:
Yes, I was thinking that `hasChangesSinceCheckpoint(checkpoint)` could be on TextBuffer, but I haven't dared propose PRs on core just yet.

`getChangesSinceCheckpoint(checkpoint)` would be even more useful – then we might proceed (maybe with `Patch`) to extract the last added character and undo only that somehow. Plus it might be useful further on, for handling undo and redo of insert operations.

Would you be willing to merge this one until TextBuffer can provide anything like the above? The current behaviour of backspace in replace mode, without this PR, is too buggy.

lee-dohm:

Issue #750, opened 2015-07-08
Change command mode cursor to default


smaili:
Right now, when in command mode or insert mode, the cursor is always `text`.  I've tried adding this selector:

```
atom-text-editor.vim-mode.command-mode {
    cursor: default;
}
```

but unfortunately that is not working.

bronson:
Just screwing around with the inspector, this seems to work:

```css
atom-text-editor.vim-mode.command-mode::shadow .editor-contents--private {
    cursor: default;
}
```

Dunno if that's anywhere near correct.

msmaili:
You're absolutely right, it does seem to work.  Although I do agree the selector isn't the most pretty :)

Is there a chance to push this as default behavior?  I think intuitively, when the editor is in command mode, it doesn't really make sense to show the cursor as text.  There's been a few times I first look to the mouse cursor and not the blinking cursor, and because it shows the text cursor I think I'm in insert mode.

bronson:
I say, create a pull request and find out.  :)

bronson:
Pretty sure this will be the idiomatic way to do this:

```less
atom-text-editor.vim-mode.command-mode
{
  &::shadow,
  &
  {
    .is-focused {
      cursor: default;
    }
  }
}
```

Think you could create a pull request that adds that to styles/vim-mode.less?

bronson:
But that, of course, fails when hovering the cursor over an unfocused pane.

`.editor-contents--private` is the only selector I can find that allows us to override the default.  So here's my most recent guess:

```less
atom-text-editor.vim-mode.command-mode
{
  &::shadow,
  &
  {
    .editor-contents--private {
      cursor: default;
    }
  }
}
```

smaili:
Just updated and it seems that `command-mode` has been renamed to `normal-mode`.

bronson:
Yep check out #811. :)

lee-dohm:
As stated in [the README](https://github.com/atom/vim-mode), this package is no longer maintained and is deprecated. We recommend that people use the [vim-mode-plus package](https://atom.io/packages/vim-mode-plus) instead. Because of this, we are [archiving this repository](https://help.github.com/articles/about-archiving-repositories/) and closing all issues and pull requests. Thanks very much for your support and contributions!

PR #811, opened 2015-08-01
use arrow mouse cursor when in normal mode

bronson:
Seems nice.  Comes from https://github.com/atom/vim-mode/issues/750


claudiodsf:
But text can still be selected when in normal mode, so the text cursor should be used.
This is, for instance, the case for any web browser: web pages are (normally) read-only, yet there is a text cursor when text can be selected.

bronson:
Yep, I agree.  Closing.

PR #812, opened 2015-08-02
Add visual feedback for search command ('/','?')

po1o:
When doing a search, display the command '/' or '?' instead of
nothing. This gives the user a visual feedback about what he is
doing.
This is achieved by adding dynamically some classes to the
element container. The CSS class selectors associated display
a content ('/' or '?') before the container.


maxbrunsfeld:
Interesting solution! Can you post screenshots or GIFs?

po1o:
vim-mode

maxbrunsfeld:
I love it.

PR #813, opened 2015-08-02
WIP Fix `dd` operation on  last line.

### LINKED ISSUE #801: `dd` line deletion on last line
Couple issues when trying to remove the last line via `dd`.

1) If the last line is completely blank and `dd` is pressed, nothing happens.

2) If the last line is not blank and `dd` is pressed, the line's contents are removed but the line itself is not.

mleeds95:
This should close #801 but I'm not sure it's the cleanest way to do it.

### DIFF
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -24,7 +24,7 @@ class Motion
         @moveSelectionInclusively(selection, count, options)
       else
         @moveSelection(selection, count, options)
-      not selection.isEmpty()
+      @isEndOfFile(selection.cursor) or not selection.isEmpty()
 
     @editor.mergeCursors()
     @editor.mergeIntersectingSelections()
@@ -106,6 +106,15 @@ class Motion
 
   isInclusive: ->
     @vimState.mode is 'visual' or @operatesInclusively
+
+  isBeginningOfFile: (cursor) ->
+    cur = cursor.getBufferPosition()
+    not cur.row and not cur.column
+
+  isEndOfFile: (cursor) ->
+    cur = cursor.getBufferPosition()
+    eof = @editor.getEofBufferPosition()
+    cur.row is eof.row and cur.column is eof.column
 
 class CurrentSelection extends Motion
   constructor: (@editor, @vimState) ->
@@ -192,10 +201,6 @@ class MoveToPreviousWholeWord extends Motion
     char = cursor.getCurrentWordPrefix().slice(-1)
     AllWhitespace.test(char)
 
-  isBeginningOfFile: (cursor) ->
-    cur = cursor.getBufferPosition()
-    not cur.row and not cur.column
-
 class MoveToNextWord extends Motion
   wordRegex: null
   operatesInclusively: false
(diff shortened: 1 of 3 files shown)


lib/operators/general-operators.coffee
```diff
@@ -88,9 +88,13 @@ class Delete extends Operator
       @setTextRegister(@register, @editor.getSelectedText())
       @editor.transact =>
         for selection in @editor.getSelections()
-          selection.deleteSelectedText()
+          if @motion.isLinewise?() and selection.getText() is @editor.lineTextForBufferRow(@editor.getLastBufferRow())
```
mleeds95: This doesn't work for a multi-line selection that contains the last line, but I have an idea of how to fix it.

mleeds95:
Actually maybe #801 isn't a bug we should fix at all. Atom renders a newline character at the end of a file as an empty row while Vim doesn't show the row at all, but I think the current behavior matches in terms of what gets written to the disk. If this PR were merged Atom would create files that end without newline characters.

lib/motions/general-motions.coffee
```diff
@@ -107,6 +107,15 @@ class Motion
   isInclusive: ->
     @vimState.mode is 'visual' or @operatesInclusively
 
+  isBeginningOfFile: (cursor) ->
+    cur = cursor.getBufferPosition()
+    not cur.row and not cur.column
```
jacekkopecky: how about `cur.equals([0,0])`?

lib/motions/general-motions.coffee
```diff
@@ -107,6 +107,15 @@ class Motion
...
+
+  isEndOfFile: (cursor) ->
+    cur = cursor.getBufferPosition()
+    eof = @editor.getEofBufferPosition()
+    cur.row is eof.row and cur.column is eof.column
```
jacekkopecky: how about `cur.equals(eof)` ?

jacekkopecky:
In the specs I see that you expect `dj` on the last line to delete the last line - but that isn't how VIM behaves, and I don't find it intuitive. Is that an intended change?

mleeds95:
You're right but as I said in my last comment this whole PR diverges from Vim's behavior in terms of what gets written to the disk.

Trailing Newline Behavior

jacekkopecky:
Yes, but that's more of a display issue - VIM chooses not to show a line
after the last \n, while Atom shows it. Both will, by default, save a \n
after the last character.

mleeds95:
@jacekkopecky I think I was able to get Atom to write a file without a newline at the end yesterday but I can't reproduce that now so I guess you're right.

mleeds95:
As for the behavior of `dj` it makes sense to me but I guess we should stick with Vim's behavior.

lib/motions/general-motions.coffee
mleeds95: I don't think `Point.equals` is defined and `is` doesn't work because the values are stored in `.row` and `.column`.

lib/motions/general-motions.coffee
mleeds95: `cur.equals(eof)` and `cur is eof` didn't seem to work for me.

jacekkopecky: Sorry, my bad, it's called isEqual in Point:
https://github.com/atom/text-buffer/blob/master/src/point.coffee#L201

PR #814, opened 2015-08-04
Fix #744

### LINKED ISSUE #744: Linewise visual select incorrectly moves cursor
To reproduce: go to the middle of a line, hit `V` to start linewise visual select, and hit `V` again to stop

What happens: the cursor jumps to the end of the line.

What should happen: the cursor remains where it originally was.

This was originally found in https://github.com/atom/vim-mode/pull/476

t9md:
Its me originally add submode shift within visual mode, at that time, I skipped to cover following case.
So original cursor position wasn't preserved.
- Entering linewise visual-mode (from normal mode) by 'V'
- Exiting linewise visual-mode.

This PR save/restore original cursor position in above entering/exiting scenario.

### DIFF
--- a/lib/operators/general-operators.coffee
+++ b/lib/operators/general-operators.coffee
@@ -194,7 +194,7 @@ class Yank extends Operator
     @setTextRegister(@register, text)
 
     @editor.setSelectedBufferRanges(newPositions.map (p) -> new Range(p, p))
-    @vimState.activateNormalMode()
+    @vimState.activateNormalMode(restoreColumn: false)
 
 #
 # It combines the current line with the following line.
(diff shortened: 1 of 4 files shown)


t9md:
I overlooked to cover `V` and `Esc` scenario. will cover soon

t9md:
To conform with Vim's convention and the spec based on that, I added `restoreColumn` argument for `Yank` motion and `Indent`.
By setting it `false`, original column could be restored after yank, I like this behavior than original Vim's one.
But need discussion, and topic for another PR.

jacekkopecky:
Can you please add some specs that would demonstrate what's fixed here? Best specs that fail without your changes, and pass with them.

t9md:
I verified spec fail without this change, and pass with this change.

PR #818, opened 2015-08-07
Fix cc on only line

jacekkopecky:
On the only line in a file, `cc` should not add a `\n`.
Also, finished the implementation of SubstituteLine with Change (alias `S` to `cc` as it should be).


maxbrunsfeld:
Great!

PR #819, opened 2015-08-08
add introspection and command to generate report

t9md:
This PR introduce new `generate-introspection-report` command which generate introspection report as markdown format to untitled buffer.
- [sample output](https://gist.github.com/t9md/cff748921480e7bf2ac5)

This is not enhancement to Vim feature, this introspection command sometime useful for vim-mode developer to get overview of each TextObject, Operator, Motion.

This introspection parse `Function.toString()` to find instance variable assignment and `super` call.
Not 100% accurate(parsing JavaScript function written in CoffeScript), but useful.

### DIFF
--- a/lib/introspection.coffee
+++ b/lib/introspection.coffee
@@ -1,0 +1,183 @@
+util = require('util')
+_ = require 'underscore-plus'
+
+extractBetween = (str, s1, s2) ->
+  str.substring(str.indexOf(s1)+1, str.lastIndexOf(s2))
+
+getParent = (obj) ->
+  obj.__super__?.constructor
+
+getAncestors = (obj) ->
+  ancestors = []
+  ancestors.push (current=obj)
+  while current = getParent(current)
+    ancestors.push current
+  ancestors
+
+inspectFunction = (fn, name) ->
+  # Calling super in the overridden constructor() function.
+  #  Case-1: No override.
+  #  CoffeeScript Source: N/A
+  #  Compiled JavaScript: return C1.__super__.constructor.apply(this, arguments);
+  #
+  #  Case-2: super without parentheses.
+  #  CoffeeScript Source: super
+  #  Compiled JavaScript: C1.__super__.constructor.apply(this, arguments);
+  #
+  #  Case-3: super with explicit argument.
+  #  CoffeeScript Source: super(a1)
+  #  Compiled JavaScript: C1.__super__.constructor.call(this, a1);
+  superBase = _.escapeRegExp("#{fn.name}.__super__.#{name}")
+  superAsIs = superBase + _.escapeRegExp(".apply(this, arguments);") # Case-2
+  defaultConstructor = '^return '+  superAsIs # Case-1
+  superWithModify = superBase + '\\.call\\((.*)\\)' # Case-3
+
+  fnString = fn.toString()
+  fnBody = extractBetween(fnString, '{', '}').split("\n").map (e) -> e.trim()
+
+  # Extract arguments from fnString. e.g. function(a1, a1){} -> ['a1', 'a2'].
+  fnArgs = fnString.split("\n")[0].match(/\((.*)\)/)[1].split(/,\s*/g)
+
+  # Replace ['arg1', 'arg2'] to ['@arg1', '@arg2'].
+  # Only when instance variable assignment statement was found.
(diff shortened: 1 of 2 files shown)


jacekkopecky:
Cool! The generic reflection part feels like it could even be a separate
NPM package.

Please add a date to the generated report, and vim-mode package version,
and if possible also current git commit ID, so it's easy to see if the
report is up to date.

Would it be possible/useful to output as the first part a subclass tree
rather than an unordered list of subclass relationships?

t9md:
>Please add a date to the generated report, and vim-mode package version,
and if possible also current git commit ID, so it's easy to see if the
report is up to date.

Will do, I'm not sure for git commit ID, will try.
Currently the generated file will written to under `~/github/vim-mode/docs`, which will convenient to update docs without moving file.  
But if we add time, version, and commit ID, we'd better to write generated file to temporary file.

> Would it be possible/useful to output as the first part a subclass tree
rather than an unordered list of subclass relationships?

I couldn't understand above, how does it like? I want some example.

t9md:
Now removed TOM-report.md itself from docs/ dir. since its now written under tmpdir.
If maintainer need to include report, he need to manually move under docs.

By the way, if this reflection utility be merged, I can add realtime opStack observation feature.  
Its also helpful to new developer understand how vim-mode works whey they try to add new TOM.

jacekkopecky:
Regarding hierarchy: currently, your output is 

- [Input](#input)
- [OperatorError](#operatorerror)
- [Operator](#operator)
- [OperatorWithInput < Operator](#operatorwithinput--operator)
- [Select < Operator](#select--operator)
- [Delete < Operator](#delete--operator)

and I think it would be useful to have it sorted alphabetically and nested by subclassing like this

- [Input](#input)
- [Operator](#operator)
  - [Delete](#delete--operator)
  - [Increase](#increase--operator)
    - [Decrease](#decrease--increase--operator)
  - [Join](#join--operator)
  - [LowerCase](#lowercase--operator)
  - [OperatorWithInput](#operatorwithinput--operator)
    - [Mark](#mark--operatorwithinput--operator)
  - [Repeat](#repeat--operator)

jacekkopecky:
The more and more I'm looking at this, the more it looks like it could be an Atom package independent of vim-mode, which would take the current project's coffee files and run this type of introspection on their exports.

Also, instead of generating a file, how about opening a new text editor and putting the content there? Then the user can choose where to save it, if at all.

t9md:
Its good idea to save non-named buiffer and let user choose where to save.
For orders hiararchy its possible but I don't have such motivation currently.

The independend Package idea is good, maybe me or other person introduce in future.

My motivation is to introduce opStack ovservation   feature on top of this introspection helper utility function  to help developer.

Its already very useful to understand where to be fixed while refactoring in my refactor ecoeriment. I can directly spot what class operator is operate on when some operator throw error.

t9md:
Updated first comment to fit to current implementation.

Here is the update summary
- Change command name `generate-reflection-report` to  `generate-introspection-report`.
- Now table of content reflect inheritance hierarchy.
- Most of code moved from `vim-state.cofee` to `introspection.coffee`.

maxbrunsfeld:
This is a really cool technique, but I don't think this code belongs inside of `vim-mode`. Maybe it should be a separate `npm` module, like @jacekkopecky said. It could be an executable, for example, that you run with any given node.js code, like this:

```sh
$ generate-introspection-report ~/.atom/packages/vim-mode/lib/*.coffee
```

t9md:
Your example seem that command is executed separate process to statically analyze coffee source.
But my code is for **dynamically** introspect within vim-mode.

The main purpose of this addition is to add `osStack` observation(=debugging) feature to vim-mode on top of this introspection function.
Here is the example output of opStack observation and explanation of how operation stack works.  
[How Operation Stack works(for current official vim-mode)](https://github.com/t9md/vim-mode#how-operation-stack-worksfor-current-official-vim-mode)

Its very useful to developer to adding or refactoring vim-mode.  

Also this introspection report can be used to explain overall change between large commit by comparing to past report.
Here is example when I removed [MotionWithInput](https://github.com/t9md/vim-mode/commit/0c7a4185ff4974ffe316459d8a98d2764057198d) intermediate class in my vim-mode folk.

Issue #755, opened 2015-07-09
vim-mode icon cannot correctly display which status(command or insert) it is


YukariyoRyu:
vim-mode icon cannot correctly display which status(command or insert) it is while editing muliti-texts. which will make confused while editing. Please solve the issue. Thank you atom team.
![screen shot 2015-07-09 at 14 17 07](https://cloud.githubusercontent.com/assets/7519371/8589144/68381f2a-2645-11e5-8708-a35a42f522f4.png)

bronson:
Strange.  Is your insert cursor always a block, or is it just sometimes?  Can you give some hints on how to reproduce this?

Do you have anything in your styles.less that's screwing it up?  Does it happen when starting in safe mode?

YukariyoRyu:
The status is that while I open a file firstly in view mode. Then I insert "i" to change to the edit mode. And so on I open another file which also in edit mode as the icon views on right bottom. But the cursor is a block not a insert-line. It appears every time when I open more than one file. I just install vim mode. I didn't install any theme .Ah my OS is Mac OS X 10.10.4 .
Thank you.

evanjbowling:
I was able to reproduce this issue in *two* different scenarios:

### Scenario A
*file in 'Normal' mode is labeled as 'Insert' mode*
1. Open file 1
2. Open file 2
3. Tab to file 1
4. Press 'i' for insert mode on file 1
5. Tab to file 2

#### Observation
Mode displayed incorrectly as 'Insert' on file 2. Block cursor still displayed as shown in screenshot above. Interactions (e.g. press 'i') prove that file is still in 'Normal' mode. Entering 'Insert' mode corrects the issue.

### Scenario B
*file in 'Insert' mode labeled as 'Normal' mode*
1. Open file 1
2. Press 'i' for insert mode on file 1
3. Open file 2
4. Press 'i' for insert mode on file 2
5. Tab to file 1
6. Press 'ESC' for normal mode on file 1
7. Tab to file 2

#### Observation
Mode displayed incorrectly as 'Normal' on file 2. Insert cursor still displayed.

#### Best Guess
These scenarios seem to apply to all other text files opened. Perhaps on the mode change of a single file, the entire array of text files currently open is being processed by mistake.

jacekkopecky:
It seems that the status bar is not updated when the user switches to a
different tab/pane etc. Should be an easy fix. Thanks for the very good
issue report.

jacekkopecky:
[...]

PR #823, opened 2015-08-13
make status bar content follow the current text editor

### LINKED ISSUE #567: Command / Insert indicator not updated when changing tabs
Steps to reproduce:

* Activate "Start In Insert Mode" in package settings
* Open two files in tabs (indicator says "Insert")
* Press [Esc] to enter command mode (indicator says "Command")
* Press [Ctrl]+[Tab] to switch tabs (indicator says "Command")

The indicator in the status bar still says "Command", but the cursor and the rest of the behavior is like in insert mode. Also pressing [Esc] changes to command mode cursor and behavior.

jacekkopecky:
Fixes #567 and #755.
I'm struggling with how to test this, though – guidance welcome. :smiley:

### DIFF
--- a/lib/status-bar-manager.coffee
+++ b/lib/status-bar-manager.coffee
@@ -1,5 +1,3 @@
-{CompositeDisposable} = require 'event-kit'
-
 ContentsByMode =
   'insert': ["status-bar-vim-mode-insert", "Insert"]
   'insert.replace': ["status-bar-vim-mode-insert", "Replace"]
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
I actually think it's ok not to test this.

Would you mind moving this new code up into the package's main module? Specifically, I'd put the call to `atom.workspace.onDidChangeActivePaneItem` right [here](https://github.com/atom/vim-mode/blob/eb12ab4b1467d9370894c7d0ae9fc45bb4a2d508/lib/vim-mode.coffee#L30). I think it would be slightly simpler that way, because the dependency relationship between the package's main module and the `StatusBarManager` would remain unidirectional.

jacekkopecky:
This is better, thank you. A little change remains in the status bar manager. Sorry about the failed build.

maxbrunsfeld:
:+1:

PR #825, opened 2015-08-15
Add last insert/yank marks

romgrk:
Hello,

This PR implements the `[` and `]` marks. 
Do not hesitate to propose changes in implementation; I wrote it as I felt was better, but it's my first PR on this package, I'll be happy to have your feedback.

Basically, it adds `VimState::setLastInsertMarks`, and calls it when `Put` and `Insert` operators are done.

Adds `::getPreviousCharPosition` -> get the Point position of the char preceding the passed position.
The reason why is because vim's `(backtick)]` leaves the cursor on the last inserted character, not after.

Documentation: http://vimhelp.appspot.com/motion.txt.html#%27%5B

### DIFF
--- a/lib/operators/input.coffee
+++ b/lib/operators/input.coffee
@@ -12,13 +12,15 @@ class Insert extends Operator
   isComplete: -> @standalone or super
 
   confirmChanges: (changes) ->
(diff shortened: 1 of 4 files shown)


lee-dohm:

Issue #826, opened 2015-08-16
Can't select last character of file using cmd-a or the mouse


DylanBruzenak:
Noticed this since the last update. 

Enter command mode and drag from the end of the file

OR 

hit cmd-a when in command mode

The last character of the file is not selected.

Using vG works. Also if you enter insert mode you can use the mouse and cmd-a as normal, which points at this being a bug in vim-mode.

jacekkopecky:
Yes, #806, working on it, sorry.

DylanBruzenak:
Duplicate of https://github.com/atom/vim-mode/issues/806

PR #827, opened 2015-08-17
Fix 806 regression

### LINKED ISSUE #806: Last character on line is skipped when selecting upward
I think there's an issue with #770.

When I'm in normal mode, and I click to the right of a line and drag upwards, the very last character on the line gets skipped.

bug3

gvim handles this correctly.

jacekkopecky:
Fixes #806 and #826 and #829.
I don't know how to add specs for the mouse movements but they are fixed too. @maxbrunsfeld please advise if you would like those specs and know how to do them.

### DIFF
--- a/lib/vim-state.coffee
+++ b/lib/vim-state.coffee
@@ -31,10 +31,13 @@ class VimState
     @marks = {}
     @subscriptions.add @editor.onDidDestroy => @destroy()
 
-    debouncedCheckSelections = _.debounce(@checkSelections, 0)
     guardedCheckSelections = =>
       return if @processing
-      debouncedCheckSelections()
+      @processing = true
+      @editorElement.component.requestAnimationFrame =>
+        @processing = false
+        @checkSelections()
+
     @subscriptions.add @editor.onDidChangeSelectionRange guardedCheckSelections
     @subscriptions.add @editor.onDidChangeCursorPosition guardedCheckSelections
     @subscriptions.add @editor.onDidAddCursor guardedCheckSelections
--- a/lib/vim-state.coffee
+++ b/lib/vim-state.coffee
@@ -689,6 +689,8 @@ class VimState
         @processing = false
       cursor.goalColumn = goalColumn
 
+    @editor.mergeCursors()
+
 # This uses private APIs and may break if TextBuffer is refactored.
 # Package authors - copy and paste this code at your own risk.
 getChangesSinceCheckpoint = (buffer, checkpoint) ->
--- a/lib/vim-state.coffee
+++ b/lib/vim-state.coffee
@@ -661,8 +661,6 @@ class VimState
 
   # Private: ensure the cursor stays within the line as appropriate
   ensureCursorsWithinLine: =>
-    return if @mode isnt 'normal'
-
     for cursor in @editor.getCursors()
       {goalColumn} = cursor
       if cursor.isAtEndOfLine() and not cursor.isAtBeginningOfLine()
--- a/lib/vim-state.coffee
+++ b/lib/vim-state.coffee
@@ -32,9 +32,12 @@ class VimState
     @subscriptions.add @editor.onDidDestroy => @destroy()
 
     debouncedCheckSelections = _.debounce(@checkSelections, 0)
-    @subscriptions.add @editor.onDidChangeSelectionRange debouncedCheckSelections
-    @subscriptions.add @editor.onDidChangeCursorPosition debouncedCheckSelections
-    @subscriptions.add @editor.onDidAddCursor debouncedCheckSelections
+    guardedCheckSelections = =>
+      return if @processing
+      debouncedCheckSelections()
+    @subscriptions.add @editor.onDidChangeSelectionRange guardedCheckSelections
+    @subscriptions.add @editor.onDidChangeCursorPosition guardedCheckSelections
+    @subscriptions.add @editor.onDidAddCursor guardedCheckSelections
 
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
@jacekkopecky Could you explain why we need specific code for handling mouse dragging, rather than treating it like any other means of updating selections?

I thought the problem was just that we prevent cursors from moving to the end of the line, even if they have non-empty selections.

jacekkopecky:
Here's what I think happens on mouse dragging in normal mode: 
 1. `mouseDown` after the end of the line, cursor moves to the end of the line and there's no selection
 2. `mouseMove`, cursor moves and selection is started and grows
The issue is that without watching the mouse, we don't know in step 1 that we shouldn't move the cursor - hence the behaviour shown in the very helpful animated gif in #806.

maxbrunsfeld:
@jacekkopecky Can you verify that this fixes the `insert-newline-above` and `join-lines` commands in normal mode? I would think that there would still be problems, because other commands outside of `vim-mode` might need to move the cursor to the end of the line, but won't be able to set the `@processing` flag. I could be wrong though; let me know.

jacekkopecky:
@maxbrunsfeld yes, I've just verified it. 

The `@processing` flag is there so that vim-mode can put the cursor immediately where it belongs after *vim-mode commands*, but it doesn't reposition the cursor while those commands are working. 

For *non-vim-mode commands*, we notice something has happened with the `onDidChangeCursorPosition` handler, but we only do something about the cursor position after the current tick (that's the `_.debounce(..., 0)` so the foreign commands are done by then.

jacekkopecky:
it means, though, that a command like `cmd-<right>` puts the cursor at the end of the line, then drawing happens, and only then the cursor is put where it belongs and redrawn. `$` doesn't suffer this issue, and vim-mode could very easily have `cmd-<right>` mapped to the same command as `$`.

jacekkopecky:
@maxbrunsfeld the last comment removes the short cursor flicker on native commands such as `cmd-<right>`. Still uncertain?

maxbrunsfeld:
Hooking into the `TextEditorComponent's` rendering cycle is an interesting solution, but it's reaching pretty far into private code, and it creates a coupling between the `VimState` and the rendering layer that I'd like to avoid.

I think that instead, we should hook into the lifecycle of Atom's commands. Basically, we'll only need to adjust the cursor position at two times: after mouse dragging ends (exactly the way you have it now), and after each command runs. Then I don't think we'll need to use `onDidChangeSelectionRange`, `onDidChangeCursorPosition`, or `onDidAddCursor` to adjust the cursor position, and I don't think we'll need the `@processing` variable on `VimState`.

In the above `atom/atom` pull request, I've added a method `atom.commands.onDidDispatch` that we can use for this. This will be available in Atom v1.0.8.

jacekkopecky:
Because we don't have a good `currentTarget` in the `did-dispatch` event (see https://github.com/atom/atom/pull/8496/files#diff-22aaebf7af3d31bbf37d33248ec7af1cR253) we won't know that the event was for our editor. We might run the cursor position and selection check for every editor on every command, but that doesn't feel right.

maxbrunsfeld:
I did a little work on this, and it seemed like checking `event.target is @editor` was sufficient.

jacekkopecky:
got it, that should work, thanks

jacekkopecky:
I now got rid of all the cursor/selection change listeners (which required a few spec changes), but it works now (with my recent enough clone of Atom, anyway).

lib/vim-state.coffee
```diff
@@ -661,15 +650,26 @@ class VimState
     text = @getRegister(name)?.text
     @editor.insertText(text) if text?
 
-  ensureCursorIsWithinLine: (cursor) =>
-    return if @processing or @mode isnt 'normal'
+  # Private: ensure the mode follows the state of selections
+  checkSelections: =>
+    return unless @editor?
+    if @editor.getSelections().every((selection) -> selection.isEmpty())
+      @ensureCursorsWithinLine() if @mode is 'normal'
```
maxbrunsfeld: It looks like you can remove the `if @mode is 'normal'` here, since `::ensureCursorsWithinLine` internally guards that the mode is normal.

maxbrunsfeld:
:heart: This is looking really good. Really excited to get this :ship:ed once Atom 1.0.8 is out.

lib/vim-state.coffee
jacekkopecky: quite right; in fact, now it's the check in `ensureCursorIsWithinLine` that's no longer necessary so I removed that one instead.

chibicode:
@maxbrunsfeld @jacekkopecky I've been following this thread for a while. Thank you for your hard work!

jacekkopecky:
Thank you for your kind words!

Oops, just remembered that I have a check for existence of onDidDispatch
that shouldn't be there; can remove it in an hour.

maxbrunsfeld:
Actually @jacekkopecky I would say leave that in there. That way, users w/ old versions of Atom won't get exceptions when they upgrade vim-mode; they just won't have the cursor adjusted when they run non-vim-mode commands.

jacekkopecky:
Hmmm, I'll have to check the amount of breakage that causes in 1.0.7. If
it's usable, you're right.

jacekkopecky:
Yup, seems usable all right

maxbrunsfeld:
:sparkles:

chibicode:
Yay! Thank you!

PR #831, opened 2015-08-19
Suppress beeps in specs

jacekkopecky:
Running specs shouldn’t beep if the beeps are expected. The patch contains commented code that, if uncommented, will print out messages when beep is called and not expected `toHaveBeenCalled()`.

### DIFF
--- a/spec/motions-spec.coffee
+++ b/spec/motions-spec.coffee
@@ -994,6 +994,12 @@ describe "Motions", ->
       vimState.globalVimState.currentSearch = {}
 
     describe "as a motion", ->
+      it "beeps when repeating nonexistent last search", ->
+        keydown '/'
+        submitNormalModeInputText ''
+        expect(editor.getCursorBufferPosition()).toEqual [0, 0]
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
Instead of having that commented-out code, I think it'd be better to explicitly assert `expect(atom.beep).not.toHaveBeenCalled()` when testing commands that *sometimes* beep, like `n`, `ctrl-a`, etc.

jacekkopecky:
Good point, added appropriate expectations of not beeping. 
As for the commented code - should I just move the spy on beeping to every test that uses it? This way, specs would start beeping again if we add beeps in other places and do not test for them.

maxbrunsfeld:
No, I think the spy on beep in spec-helper is good. I would just remove the commented out code. I'd prefer to avoid this kind of auxiliary code in this repo; since it isn't automatically exercised by CI, it can tend to become invalid as the code changes.

jacekkopecky:
OK, very good point. Removed.

maxbrunsfeld:
:heart:

PR #836, opened 2015-08-22
Fix dd so it deletes last empty line. Fix #801

### LINKED ISSUE #801: `dd` line deletion on last line

octref:
Fix #801

### DIFF
--- a/lib/operators/general-operators.coffee
+++ b/lib/operators/general-operators.coffee
@@ -94,6 +94,11 @@ class Delete extends Operator
           cursor.skipLeadingWhitespace()
         else
           cursor.moveLeft() if cursor.isAtEndOfLine() and not cursor.isAtBeginningOfLine()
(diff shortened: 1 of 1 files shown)


octref:
Haven't written test yet. Would like to know if this is the right approach before I go too deep.

mleeds95:
Feel free to copy the tests from https://github.com/mleeds95/vim-mode/commit/0c4d1ace56aa658b59573ea7eaf84ad3cb6e903a. I never got around to fixing #813.

octref:
I'll be busy for a while. I guess I'll come back when I have time and start a new PR from scratch.

PR #837, opened 2015-08-23
Fix multi-cursor indent

jacekkopecky:
Indent, e.g. `>>`, should not collapse multiple cursors. This PR fixes it.

Here's steps to reproduce the problem:
 1. add a few cursors (e.g. with `cmd-click`)
 2. press `>>`, each line gets indented
 3. observe that cursors other than the last one are gone


maxbrunsfeld:
Awesome thanks for spotting that.

PR #839, opened 2015-08-24
Fix default counts

### LINKED ISSUE #838: gg=G (format all code)
I've gotten used to formatting code in this way.

I think I originally saw it on this [vim wiki page][1]:

[1]: http://vim.wikia.com/wiki/Fix_indentation

> In normal mode, typing gg=G will reindent the entire file. This is a special case; = is an operator. Just like d or y, it will act on any text that you move over with a cursor motion command. In this case, gg positions the cursor on the first line, then =G re-indents from the current cursor position to the end of the buffer.

I'm guessing it'll be easy to fix because it does work in visual mode ggvG=<kbd>esc</kbd>, which also means it's a low priority.

jacekkopecky:
Fixes #838.
The issue was that some operators (including `=`) had an unnecessary default `count` of 1, but the `G` motion is exceptional in that its default is different from 1. The specs never tested the combination of G with the offending operators, now they do.

### DIFF
--- a/lib/operators/general-operators.coffee
+++ b/lib/operators/general-operators.coffee
@@ -103,7 +103,7 @@ class Delete extends Operator
 class ToggleCase extends Operator
   constructor: (@editor, @vimState, {@complete}={}) ->
 
-  execute: (count=1) ->
+  execute: (count) ->
     if @motion?
       if _.contains(@motion.select(count), true)
         @editor.replaceSelectedText {}, (text) ->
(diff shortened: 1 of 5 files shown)


lib/operators/general-operators.coffee
```diff
@@ -119,7 +119,7 @@ class ToggleCase extends Operator
         for cursor in @editor.getCursors()
           point = cursor.getBufferPosition()
           lineLength = @editor.lineTextForBufferRow(point.row).length
-          cursorCount = Math.min(count, lineLength - point.column)
+          cursorCount = Math.min(count ? 1, lineLength - point.column)
```
maxbrunsfeld: Why do this, rather than use a default argument value like before?

jacekkopecky: because here, `count` is also used in `@motion.select(count)` where a default was the problem.

PR #844, opened 2015-08-26
make text objects expand the selection, not replace it

### LINKED ISSUE #754: iw in visual select destroys your selection
- start with a file that contains a few words.  move to a word and hit `v`.
- hit `w` to select until the next word
- hit `iw` to add that word to the selection

In Vim, the selection grows to include that word.  In Atom, your selection gets destroyed and only that word ends up selected.

jacekkopecky:
Fixes #754.


maxbrunsfeld:
Nice!

PR #846, opened 2015-08-26
ignore input during IME composition in VimNormalModeInputElement

### LINKED ISSUE #791: Replace (r) does not work with accented characters on OSX
In OSX you can type accented characters (éåñ) etc directly from keyboard. 
To type é, you first press alt+e, then e. Replacing the character with (r) should allow to type in the accented version of the character, but it doesn't.

Example:

I want to type México, so I type Mexico and then want to replace e with é. So I place the cursor above e and press (r) and then alt+e,e. 

Expected result: México
Actual result: M´xico.

jacekkopecky:
Fixes #791.
Adapted from [atom/autocomplete-plus](https://github.com/atom/autocomplete-plus/blob/90ae1f3ac5c2a7a46bacf3c1b7c891e1902aad42/lib/autocomplete-manager.coffee#L76)'s workaround for atom/atom#5414.
Specs adapted from atom/atom#8526 .

### DIFF
--- a/lib/view-models/vim-normal-mode-input-element.coffee
+++ b/lib/view-models/vim-normal-mode-input-element.coffee
@@ -28,8 +28,11 @@ class VimNormalModeInputElement extends HTMLDivElement
 
   handleEvents: ->
     if @singleChar?
+      compositing = false
       @editorElement.getModel().getBuffer().onDidChange (e) =>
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
:+1:

Issue #521, opened 2015-02-12
Uncaught TypeError: Cannot read property 'length' of undefined


tali:
[Enter steps to reproduce below:]

I cannot reproduce it any more. I was playing with vim-mode and markdown-preview and updated to the latest Atom (now 0.179.0) and vim-mode (now 0.31.0). Maybe it was related to some inconsistent state caused by upgrading atom or the vim-mode package.

What I remember:
1. open markdown file and open markdown-preview
2. Now I have problems navigating in the text (not reacting to vim movement commands)
3. Close and reopen the markdown file
4. Now I could move again, but could not move to the last lines with vim commands
5. Clicking into the last line, the cursor got moved
6. Using vim movement commands to move within that line (h,l) triggered the exception

I hope the stack trace still helps to find some bugs :-)

**Atom Version**: 0.179.0
**System**: Mac OS X 10.10.2
**Thrown From**: [vim-mode](https://github.com/atom/vim-mode) package, v0.31.0
### Stack Trace

Uncaught TypeError: Cannot read property 'length' of undefined

```
At /Users/martin/.atom/packages/vim-mode/lib/motions/general-motions.coffee:102

TypeError: Cannot read property 'length' of undefined
  at /Users/martin/.atom/packages/vim-mode/lib/motions/general-motions.coffee:102:15
  at Object._.times (/Users/martin/.atom/packages/vim-mode/node_modules/underscore-plus/node_modules/underscore/underscore.js:1119:53)
  at MoveRight.execute (/Users/martin/.atom/packages/vim-mode/lib/motions/general-motions.coffee:103:5)
  at VimState.module.exports.VimState.processOpStack (/Users/martin/.atom/packages/vim-mode/lib/vim-state.coffee:148:15)
  at VimState.module.exports.VimState.pushOperations (/Users/martin/.atom/packages/vim-mode/lib/vim-state.coffee:141:35)
  at atom-text-editor.commands.(anonymous function) (/Users/martin/.atom/packages/vim-mode/lib/vim-state.coffee:136:34)
  at atom-text-editor.handler (/Applications/Atom.app/Contents/Resources/app/src/space-pen-extensions.js:115:32)
  at atom-text-editor.jQuery.event.dispatch (/Applications/Atom.app/Contents/Resources/app/node_modules/space-pen/vendor/jquery.js:4681:9)
[...]

PR #848, opened 2015-08-27
fix issue #521

dnasseri:
Fix issue #521

### DIFF
--- a/lib/motions/search-motion.coffee
+++ b/lib/motions/search-motion.coffee
@@ -169,6 +169,7 @@ class BracketMatchingMotion extends SearchBase
 
       if point.column < 0
         point.row--
+        return null if point.row < 0
         lineLength = @editor.lineTextForBufferRow(point.row).length
         point.column = lineLength - 1
       else if point.column >= lineLength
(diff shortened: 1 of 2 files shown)


maxbrunsfeld:
Thanks for figuring this out @dnasseri. Could you add a test for this? You should add it somewhere in [this area](https://github.com/atom/vim-mode/blob/2ab86b2942efe6fee81e3d6492ba2020e64ee499/spec/motions-spec.coffee#L1832).

dnasseri:
@maxbrunsfeld absolutely!

dnasseri:
@maxbrunsfeld I'm having a little bit of trouble with getting the spec to fail appropriately

this is the spec I wrote:

    it 'stays put', ->
      editor.setText("\n)")
      editor.setCursorScreenPosition([1, 0])
      expect(-> keydown('%')).not.toThrow();

When I comment out the fix and run `apm test` I see that the error is thrown in the console, but the test does not fail. A bit new to coffeescript. Am I missing something here? Thanks!

dnasseri:
Closing until I fix the spec.

PR #854, opened 2015-08-31
suppress unwanted bracket matching

### LINKED ISSUE #792: Auto insertion should be disabled when in replace mode (SHIFT + R)
When in replace mode (via `SHIFT + R`) entering `"` causes the current (as expected) as well as next character (unexpected) to be replaced with two `"`

This happens with and other character with a corresponding closing character; `{}`, `[]`, `'` etc. and only when atom would normally insert the closing character in insert mode.

For example, given image, pressing `SHIFT + R` then `"` you end up with:
image

The next character will replace the second `"` as expected, but the second `"` should have never been inserted.

The expected output is image

jacekkopecky:
Fixes #792, but depends on atom/bracket-matcher#176.

You'll notice that there's a lot of _to-do_s for specs which I'll be happy to add once this whole approach is approved.

To do:
- [ ] get atom/bracket-matcher#176 merged
- [ ] add specs

### DIFF
--- a/lib/insert-mode.coffee
+++ b/lib/insert-mode.coffee
@@ -1,3 +1,4 @@
+# todo add specs with bracket-matcher for these two
 copyCharacterFromAbove = (editor, vimState) ->
   editor.transact ->
     for cursor in editor.getCursors()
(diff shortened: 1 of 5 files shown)


lee-dohm:

Issue #853, opened 2015-08-31
ci'  command does not work


mushketyk:
**Version:** 0.59.0

**How to reproduce:**
1. Create a new file
2. Write: 'my-project'
3. In normal mode set cursor to character "o"
4. Enter the following command: ci'

**What happens:** nothing
**What should happen:** "my-project" should be removed and editor should enter insert mode

jacekkopecky:
@mushketyk can you please check what the Key Binding Resolver (toggled with `cmd-.` or in the Packages menu) shows when you've pressed `ci'`?

mushketyk:
@jacekkopecky Hi, I could only find bindings for "c" in Packages -> Settings View -> Show Keybindings:

mushketyk:
I also noticed that I can reproduce this bug only from time to time and if I restart Sublime everything works fine again.

jacekkopecky:
I'm especially interested in what the keybinding resolver shows when you
press c and ' please.

Also, you are restarting Atom along with Sublime, right? :smiley:

alexbooker:
I am a new `vim-mode` user and this is a limitation I have encountered as well.

The _Key Binding Resolver_ shows `i'`.

And here is the code I am trying to perform the operation on:

```elixir
ExUnit.start

defmodule MyTest do
  use ExUnit.Case
    test 'i wanna replace text in here pls' do
      assert 1 + 1 == 2
    end
end
```

alexbooker:
Is there any more information you would like from me?

I noticed that `ci{` etc. works fine.

jacekkopecky:
@alexbooker can you reliably reproduce the condition? I tried `ci'` in various places of your test name and it worked properly. And if you can reproduce it, does it also show up when you disable packages other than vim-mode and restart Atom?
 
It could also be that the user sometimes, through a distraction, spends more than a second between typing `i` and `'` – that would cancel the Atom command.

alexbooker:
> Can you reliably reproduce the condition?

@jacekkopecky Yes, I can.

> And if you can reproduce it, does it also show up when you disable packages other than vim-mode and restart Atom?

Yes.

> It could also be that the user sometimes, through a distraction, spends more than a second between typing i and ' – that would cancel the Atom command.

It *does* work with `ci{` etc. so I do not think this is the problem.

I am running Atom 1.0.11 on Windows 10 64bit. I'll try reinstalling Atom now. 

Any other questions I can answer to help solve this? Thanks.

[...]

PR #862, opened 2015-09-14
Made clause about international keyboards specific

alexbooker:
**Proposed changes:**

- The text used to read *"vim-mode has some issues with international keyboard layouts"*. The *"International"* part was too general in my opinion, so I changed it to *"non-US"* which - [in my experience](https://github.com/atom/vim-mode/issues/853) - is true.
- A solution which worked really well for *me* was to install [*keyboard-localization*](https://atom.io/packages/keyboard-localization). I *imagine* more people will find it useful - could be a good idea to mention it.

**Other:**

- When I read *"international"* I thought of countries whose native language isn't English (e.g. France and Turkey). I *definitely* thought UK and US were similar enough but it turns out UK is different enough to cause problems. I imagine that other people in the UK could get confused by this. *Maybe* it would be a good idea to update the text to specify this explicitly e.g.:

> Currently, vim-mode has some issues with non-US keyboard layouts (*including* UK keyboard layouts).

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ wasn't built in a day).
 If you want the vim ex line (for `:w`, `:s`, etc.), you can try [ex-mode](https://atom.io/packages/ex-mode)
 which works in conjuction with this plugin.
 
-Currently, vim-mode has some issues with international keyboard layouts.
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
:+1: Thanks for this, @alexbooker.

PR #866, opened 2015-09-22
Activate vim insert mode by pressing 'Insert' key.

naw:
This works in vi and vim --- might as well work in vim-mode too.


maxbrunsfeld:
:+1: Makes sense. Does our key-mapping system handle `insert` properly?

naw:
Yes. At least, pressing insert on the keyboard _does_ activate this action as expected as a result of adding the 'insert' mapping.

maxbrunsfeld:
Thanks!

Issue #872, opened 2015-09-25
Jump-to-nearest when selecting around/inside quotes includes all characters beginning at cursor


MarkusSN:
Steps to reproduce:
Given a line `var a = 'awesome text'`
Place cursor over `r` and input `vi'`
Expected selection `awesome text`
Actual selection `r a = 'awesome text`

lee-dohm:
As stated in [the README](https://github.com/atom/vim-mode), this package is no longer maintained and is deprecated. We recommend that people use the [vim-mode-plus package](https://atom.io/packages/vim-mode-plus) instead. Because of this, we are [archiving this repository](https://help.github.com/articles/about-archiving-repositories/) and closing all issues and pull requests. Thanks very much for your support and contributions!

PR #873, opened 2015-09-26
Fix jump-to-nearest for quotes text objects

MarkusSN:
This is a fix for #872.

I couldn't think of a way to distinguish between when we want to merge selections and when we want to jump-to-nearest. So, I just removed the call to mergeRanges, and replaced the tests that expected the ranges to be merged.

### DIFF
--- a/lib/text-objects.coffee
+++ b/lib/text-objects.coffee
@@ -88,7 +88,7 @@ class SelectInsideQuotes extends TextObject
         ++ start.column # skip the opening quote
         end = @findClosingQuote(start)
         if end?
-          selection.setBufferRange(mergeRanges(selection.getBufferRange(), [start, end]))
+          selection.setBufferRange([start, end])
       not selection.isEmpty()
 
 # SelectInsideBrackets and the previous class defined (SelectInsideQuotes) are
--- a/spec/text-objects-spec.coffee
+++ b/spec/text-objects-spec.coffee
@@ -453,17 +453,13 @@ describe "TextObjects", ->
       expect(editorElement.classList.contains('operator-pending-mode')).toBe(false)
       expect(editorElement.classList.contains('normal-mode')).toBe(true)
 
-    it "expands an existing selection in visual mode", ->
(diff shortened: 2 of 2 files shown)


jacekkopecky:
Is this behaviour consistent with other text objects? Can't test it today,
don't want to forget.

MarkusSN:
No, this change will mean that invoking the quotes-text-object will replace the current selection instead of expanding it, unlike eg. the paragraph- or word-text-objects.

jacekkopecky:
Thank you for clarifying. Now I see Vim is similarly inconsistent. Is it
desirable that vim-mode should replicate this behaviour?

MarkusSN:
In my opinion, yes. Being able to jump to the nearest quoted string on the same line is very handy.

jacekkopecky:
If you're not in visual mode, jumping to the nearest quoted string and selecting its contents is the default behaviour of the `vim-mode:select-inside-double-quotes` action. Since vim-mode doesn't support `gi` at the moment (nor have I seen any calls for it), perhaps we might bind `gi"` to _go to inside of double quotes_? 
For now, if your muscle memory doesn't terribly mind, you can put the following in your keymap:
```
'atom-text-editor.vim-mode.normal-mode':
  'g i "': 'vim-mode:select-inside-double-quotes'
```

MarkusSN:
Thanks for suggestion! But I'm afraid this one is heavily ingrained in muscle memory :smile:

Another thing I realized while looking at the different text object implementations is that most of them do something along these lines, i.e. relying on selection._cursor_:
```
      [...] selection.cursor.getBufferPosition())
```
The problem is that when we enter visual mode, the cursor is moved one character to the right. This means that there exists a few off-by-one bugs when selecting text objects.

If visual mode was changed so that we do not create a selection before the user actually moves the cursor, we might be able to keep the current merge logic and jump-to-nearest should start working. And also fix the off-by-one bugs.

I don

lee-dohm:

PR #877, opened 2015-09-30
Fix deprecated methods

as-cii:
This will probably fail on Travis, as the new version hasn't been released yet. However, specs pass locally :white_check_mark: and the package seems to work just fine.

The changes in this PR are quite straightforward, but I think it's fine if we :ship: this as soon as the new editor version lands. Also, we should probably bump the engine version in `package.json`.

/cc: @maxbrunsfeld for :eyes:

### DIFF
--- a/spec/spec-helper.coffee
+++ b/spec/spec-helper.coffee
@@ -19,8 +19,8 @@ getEditorElement = (callback) ->
       textEditor = e
 
   runs ->
-    element = document.createElement("atom-text-editor")
-    element.setModel(textEditor)
+    element = atom.views.getView(textEditor)
+    element.setUpdatedSynchronously(true)
     element.classList.add('vim-mode')
     statusBarManager ?= new StatusBarManager
     globalVimState ?= new GlobalVimState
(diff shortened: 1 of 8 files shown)


spec/spec-helper.coffee
```diff
@@ -58,7 +58,7 @@ keydown = (key, {element, ctrl, shift, alt, meta, raw}={}) ->
   key = "U+#{key.charCodeAt(0).toString(16)}" unless key is 'escape' or raw?
   element ||= document.activeElement
   eventArgs = [
-    true, # bubbles
+    false, # bubbles
```
as-cii: Please, note that I had to disable bubbling here because, for some reason, events were being fired twice. I guess this has something to do with attaching the element to the DOM but I haven't quite understood exactly what was causing it.

@maxbrunsfeld: I guess I can have a second look if you think it's needed, but this change didn't seem that bad either.

maxbrunsfeld:
Looks good to me. Let's merge and release this as soon as v1.1.0 is out.

as-cii:
Thanks, @maxbrunsfeld! Please, note that I have bumped also the engine version in 6165b66.

PR #879, opened 2015-09-30
Fix deprecated methods

as-cii:
Continuation of #877.


maxbrunsfeld:
Fixes https://github.com/atom/vim-mode/issues/903, I believe.

as-cii:
I think the two failing specs are red because of subpixel font scaling: changing `20` to `19` should be okay there.

maxbrunsfeld:
Ok, I was wrong about this fixing #903. Investigating...

PR #881, opened 2015-10-07
:bug: Fix spec description for ctrl-u

mattr-:
ctrl-u moves up the screen but the spec description said it moved down.


maxbrunsfeld:
:ok_hand:

PR #886, opened 2015-10-09
Implement sentence motions.

jonasws:
This PR implements the traditional `)` and `(` keybindings originally found in vim. I would greatly appreciate any comments and feedback on how I made it work with regex, as I'm certain there may be a slightly slicker way to pull it off (though I've tried to do similar to how other motions are implemented).

### DIFF
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -296,7 +296,7 @@ class MoveToNextSentence extends Motion
       eof = cursor.editor.getBuffer().getEndPosition()
       scanRange = [start, eof]
 
-      cursor.editor.scanInBufferRange /([\.!?] )|^[A-Za-z0-9]/, scanRange, ({matchText, range, stop}) ->
+      cursor.editor.scanInBufferRange /(^$)|(([\.!?] )|^[A-Za-z0-9])/, scanRange, ({matchText, range, stop}) ->
         adjustment = new Point(0, 0)
         if matchText.match /[\.!?]/
           adjustment = new Point(0, 2)
@@ -311,7 +311,7 @@ class MoveToPreviousSentence extends Motion
       bof = cursor.editor.getBuffer().getFirstPosition()
       scanRange = [bof, end]
 
-      cursor.editor.backwardsScanInBufferRange /([\.!?] )|^[A-Za-z0-9]/, scanRange, ({matchText, range, stop}) ->
+      cursor.editor.backwardsScanInBufferRange /(^$)|(([\.!?] )|^[A-Za-z0-9])/, scanRange, ({matchText, range, stop}) ->
         adjustment = new Point(0, 0)
         if matchText.match /[\.!?]/
           adjustment = new Point(0, 2)
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -289,6 +289,36 @@ class MoveToEndOfWord extends Motion
 class MoveToEndOfWholeWord extends MoveToEndOfWord
   wordRegex: WholeWordRegex
 
+class MoveToNextSentence extends Motion
+  moveCursor: (cursor, count=1) ->
+    _.times count, ->
+      start = cursor.getBufferPosition().translate new Point(0, 1)
+      eof = cursor.editor.getBuffer().getEndPosition()
+      scanRange = [start, eof]
+
+      cursor.editor.scanInBufferRange /([\.!?] )|^[A-Za-z0-9]/, scanRange, ({matchText, range, stop}) ->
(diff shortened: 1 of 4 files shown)


mattr-:
Great work @jonasws. There's still some behavior differences that we might want to look at addressing.

For example, in this repo, open `spec/motions-spec.coffee` in vim and go to line 1340. Now hit `)` to go to the next sentence. In vim, this moves you to line 1342. With this patch, Atom does nothing.

While still on line 1340 in the same file, hit `(` to back a sentence and you're taken all the way back to line 1081 where there's a `?` with a space after it in a string. Doing the same thing in Vim only takes you back to the beginning of line 1339.

The larger conclusion to draw from this is that paragraph and section boundaries don't seem to count as sentence boundaries as the Vim docs say they should. From vim's `:help sentence`:
>A sentence is defined as ending at a `.`, `!` or `?` followed by either the
end of a line, or by a space or tab.  Any number of closing `)`, `]`, `"`
and `'` characters may appear after the `.`, `!` or `?` before the spaces,
tabs or end of line.  A paragraph and section boundary is also a sentence
boundary.

Just my 2cents, as I'm not the actual maintainer of this plugin. :smiley:

jonasws:
@mattr That is great feedback! I am not an expert vim user myself, so I figured someone had something to comment on that I completely missed. I also implemented this with the purpose of it to work on files consisting of mainly text (such as Markdown and LaTex), as I didn't think it was defined by vim what would happen when pressing those keys in a code file. I'll look into what vim says to try to make it compliant.

maxbrunsfeld:
Ok well the code looks very nice. It seems like a paragraph breaks, or empty lines, should be considered sentence boundaries, but other than that, I'd love to :ship: this.

jonasws:
@maxbrunsfeld I have implemented sentence boundaries as we discussed. I have another question that we might want to address. How do we want things to work in "code files", where sentences do not occur as the normally would? I haven't been able to think of anything clever on that account.

maxbrunsfeld:
I don't think you need to change the behavior based on the filetype or context.

maxbrunsfeld:
Are you happy w/ this as is, for your use case? If so, I'm fine with merging this as long as the tests pass. We can always refine the behavior in a subsequent PR.

jonasws:
I am happy with it as it is. As you said, making it work exactly like vim, is quite complex, and should probably be done in another PR, to keep the review work simpler.

spec/motions-spec.coffee
```diff
@@ -316,6 +316,55 @@ describe "Motions", ->
...
+        keydown '('
+        expect(editor.getCursorBufferPosition()).toEqual [2, 20]
+
+        keydown '('
+        expect(editor.getCursorBufferPosition()).toEqual [2, 0]
+
+        keydown '('
+        expect(editor.getCursorBufferPosition()).toEqual [1, 0]
+
+
```
maxbrunsfeld: Could you remove this second blank line?

lib/motions/general-motions.coffee
```diff
@@ -289,6 +289,36 @@ class MoveToEndOfWord extends Motion
...
   wordRegex: WholeWordRegex
 
+class MoveToNextSentence extends Motion
+  moveCursor: (cursor, count=1) ->
+    _.times count, ->
+      start = cursor.getBufferPosition().translate new Point(0, 1)
+      eof = cursor.editor.getBuffer().getEndPosition()
+      scanRange = [start, eof]
+
+      cursor.editor.scanInBufferRange /(^$)|(([\.!?] )|^[A-Za-z0-9])/, scanRange, ({matchText, range, stop}) ->
```
maxbrunsfeld: Could you explain why `^[A-Za-z0-9]` is included here? I would we would just want to scan for sentence punctuation (`.`, `?`, `!`) and empty lines.

jonasws: I figured I wanted to follow the definition of a sentence, but you may well be  right that it's a bit overkill.

maxbrunsfeld: Oh I see; that make sense. :+1:

jonasws: Also, after some testing, I've discovered that it is necessary to search for some character (maybe just a printable non-whitespace character) in order to accomplish the "next sentence on next line" edge case. I'm still looking to see if we don't need the A-Za-z or similar character sequence, but no luck so far.

jonasws: Update on testing. Could not correctly implement the behavior when removing the letters from the regex. I'd suggest it could go as an improvement for a later PR.

maxbrunsfeld:
Thanks @jonasws!

PR #889, opened 2015-10-10
Keymap for toggle comments

jaguadoromero:
I've added a keymap for toggle comments with "g c c" shortcut like some vim plugins


maxbrunsfeld:
Hi, this is nice, but I don't think it should be built into vim-mode itself. People can add this key-binding separately from installing this package.

PR #890, opened 2015-10-11
Fix c/d/y iw/aw commands when run from the first character of a word to not include delimeters

### LINKED ISSUE #868: "ciw" doesn't respect delimiters when cursor is at beginning of word
Steps to reproduce (`|` is cursor in normal mode):

`abc.de|f` → type `ciw` → `abc.|` (expected behaviour)
`abc.|def` → type `ciw` → `abc|` (expected: `abc.|`)

Is this an isolated bug or a side effect of a larger existing issue?

benmccormick:
This fixes the current behavior where `ciw` and related commands cut off delimeters before words when on the first letter.  IE for "/foo/bar/whatever"  would become "/foo/whatever" if you navigated to the 'b' and ran `diw`

As far as I can tell this behavior is caused because `selection.expandOverWord` expands to 2 words when the selection is empty and the cursor is "between" 2 words.  Even though vim-mode uses block cursor, atom considers your cursor to be between the "/" word and the "bar" word.  

I've resolved this by explicitly selecting forward if the selection is empty and then expanding to word.  

This closes #868

### DIFF
--- a/lib/text-objects.coffee
+++ b/lib/text-objects.coffee
@@ -14,6 +14,8 @@ class TextObject
 class SelectInsideWord extends TextObject
   select: ->
     for selection in @editor.getSelections()
+      if selection.isEmpty()
+        selection.selectRight()
       selection.expandOverWord()
     [true]
 
@@ -144,6 +146,8 @@ class SelectInsideBrackets extends TextObject
 class SelectAWord extends TextObject
   select: ->
     for selection in @editor.getSelections()
+      if selection.isEmpty()
+        selection.selectRight()
       selection.expandOverWord()
       loop
         endPoint = selection.getBufferRange().end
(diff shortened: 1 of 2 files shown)


benmccormick:
@maxbrunsfeld (or anyone else) is there anything I can do to get this moving? The behavior is pretty annoying to me.

benmccormick:
@50Wliu @maxbrunsfeld is there anyone I can work with to move this along?

maxbrunsfeld:
Hey @benmccormick, sorry for the delay on this. Thanks for the fix!

benmccormick:
@maxbrunsfeld great!  No worries, I know you all have more important things in life :)  Glad it got in!

maxbrunsfeld:
This is in 0.62.

Issue #903, opened 2015-10-30
Cursor does not follow ctrl+f and ctrl+b after upgrade to Atom 1.1


damassi:

thomasjo:
Please provide more details about your setup; OS, installed packages, etc.

damassi:
Sure thing -- OSX El Capitan (pretty standard setup) and here's a dump of my plugins. Downgrading fixed the issue.

atom-alignment@0.12.1
atom-beautify@0.28.16
atom-jade@0.3.0
atom-pair@2.0.10
atom-runner@2.4.6
autocomplete-php@0.3.7
base16-eighties-dark-syntax@0.1.0
black-and-white-ui@0.1.15
bottom-dock@0.3.5
close-other-tabs@0.1.0
color-picker@2.0.13
cursor-history@0.5.5
docblockr@0.7.3
git-plus@5.4.7
highlight-selected@0.10.1
ide-flow@0.8.0
indent-guide-improved@1.4.3
language-ejs@0.2.0
language-elm@1.2.0
language-haml@0.23.4
language-haskell@1.4.12
language-javascript-jsx@0.3.7
linter@1.9.1
linter-elm-make@0.3.0
linter-eslint@3.2.0
linter-php@1.0.5
markdown-scroll-sync@0.4.2
merge-conflicts@1.3.6
nuclide-click-to-symbol@0.0.22
one-dark-ui@1.1.5
open-terminal-here@2.1.0
painless-panes@0.1.0
react-snippets@0.2.0
script@3.0.1
solo-one-dark-ui@0.1.1
symbol-gen@1.0.0
sync-settings@0.6.0
tabs-to-spaces@0.11.1
term2@0.9.21
todo-manager@0.2.6
todo-show@0.16.0
trailing-semicolon@1.1.0
turbo-javascript@1.1.5
vim-mode@0.62.0
vim-mode-visual-block@0.2.13
zentabs@0.8.6

thedaniel:
@damassi can you reproduce with safe mode? You can launch Atom in safe mode with `atom --safe` in the terminal or `Application: Open Safe` in the command panel.

thomasjo:
I was going to propose the same, but vim-mode is not a core package, so it'll be disabled when running in safe mode.

thedaniel:
Whoops!

damassi:
Yeah, my setup is nothing special. Could someone please see if paging works with 1.1?

maxbrunsfeld:
[...]

PR #906, opened 2015-11-02
Deal with not-yet-rendered scroll position changes in scrolling commands

maxbrunsfeld:
Fixes https://github.com/atom/vim-mode/issues/903

### DIFF
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -451,54 +451,50 @@ class MoveToMiddleOfScreen extends MoveToScreenLine
     height = lastScreenRow - firstScreenRow
     Math.floor(firstScreenRow + (height / 2))
 
-class ScrollKeepingCursor extends MoveToLine
-  previousFirstScreenRow: 0
-  currentFirstScreenRow: 0
+class ScrollKeepingCursor extends Motion
+  operatesLinewise: true
+  cursorRow: null
 
   constructor: (@editorElement, @vimState) ->
     super(@editorElement.getModel(), @vimState)
 
   select: (count, options) ->
-    finalDestination = @scrollScreen(count)
+    scrollTop = @scrollScreen(count)
     super(count, options)
-    @editorElement.setScrollTop(finalDestination)
+    @editorElement.setScrollTop(scrollTop)
 
   execute: (count) ->
-    finalDestination = @scrollScreen(count)
+    scrollTop = @scrollScreen(count)
     super(count)
-    @editorElement.setScrollTop(finalDestination)
+    @editorElement.setScrollTop(scrollTop)
 
-  moveCursor: (cursor, count=1) ->
-    cursor.setScreenPosition([@getDestinationRow(count), 0])
-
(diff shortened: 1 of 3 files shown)


lib/motions/general-motions.coffee
```diff
@@ -451,54 +451,50 @@ class MoveToMiddleOfScreen extends MoveToScreenLine
...
   scrollScreen: (count=1) ->
-    @previousFirstScreenRow = @editorElement.getFirstVisibleScreenRow()
-    destination = @scrollDestination(count)
-    @editorElement.setScrollTop(destination)
-    @currentFirstScreenRow = @editorElement.getFirstVisibleScreenRow()
-    destination
+    # TODO: We need to create an API for reading an editor's logical scroll
+    # position which may not yet have been flushed to the DOM. Currently, the
+    # only way to access this information is through this piece of private state.
+    currentScrollTop = @editorElement.component.presenter.pendingScrollTop ? @editorElement.getScrollTop()
```
maxbrunsfeld: /cc @nathansobo @as-cii 

In this command, we're trying to read the editor's current scroll position, and update it accordingly. The problem is that if we run this command rapidly (e.g. holding down `ctrl-f`), calls to `TextEditorElement::getScrollTop()` will not reflect the value that was previously set via `::setScrollTop()`, because a DOM update hasn't happened yet.

Can you think of an easier way to do this? If not, maybe we need to add an API on `TextEditorElement` for accessing its current *desired* scroll position (i.e. the last value passed to `::setScrollTop`).

maxbrunsfeld:
Gonna :ship: this as is for a quick fix, but hopefully we can come up with something cleaner.

as-cii:
Thanks Max for addressing this, will take a better look at this tomorrow morning. 🙏

lib/motions/general-motions.coffee
as-cii: Thanks @maxbrunsfeld for bringing this up. So, there are a couple of ways in which we could solve this. We may either:

1. Add something like a `TextEditorElement::getPendingScrollTop` API;
2. Flush pending state (such as scroll positions) synchronously whenever methods like `::getScrollTop` get called;

Personally, I would prefer 2) over 1) because it both feels more consistent with how the DOM works like and it doesn't expose an internal information which heavily depends on the update scheduling implementation. For example: what does pending mean here? In the current implementation we know that the scroll position's state is either pending or committed, but what if we started handling this differently by introducing a third state?

On the other hand, I understand the implications of having such a synchronous API: forcing reflows is not ideal, especially when the interaction with the user needs to happen repeatedly such as in this case (e.g. holding down a key). It's also true that people are already used to how the DOM behaves when reading/writing from/to it, so I feel like this wouldn't be a deal breaker. Indeed, this piece of code could be changed to make use of `requestAnimationFrame` (or our `ViewRegistry::readDocument` + `ViewRegistry::updateDocument` methods) to orchestrate DOM updates in a way that doesn't affect performance because of synchronous reflows.

What do you think?

PR #912, opened 2015-11-09
Use model-level scrolling API if available

maxbrunsfeld:
Depends on https://github.com/atom/atom/pull/9444


maxbrunsfeld:
For backwards compatibility, this uses the old code-path if if `TextEditor::setFirstVisibleScreenRow` is not defined.

PR #917, opened 2015-11-12
map | (pipe) to move-to-beginning-of-line

globau:
for vim parity map | (pipe) to move-to-beginning-of-line


maxbrunsfeld:
Thanks!

PR #922, opened 2015-11-15
Remove unneccsary commas in vim-mode.cson

jonasws:
@maxbrunsfeld Removed the commas as we discussed. Sorry for the "blunder".


maxbrunsfeld:
:ok_hand: Thanks for taking care of this @jonasws!

PR #931, opened 2015-12-05
fixes jumping with large line selections, fixes #930

### LINKED ISSUE #930: selecting with shift+v jumps strangely when scrolling with j/k when more than one screens worth of lines is selected
Doesn't happen with selecting with v
Here's what the behavior looks like: [gfy](http://gfycat.com/DevotedEmptyAracari)
gif

ssfrr:


maxbrunsfeld:
Thanks @ssfrr!

PR #940, opened 2015-12-17
Use clipboard as default register by default

### LINKED ISSUE #541: ctrl-c and ctrl-v not working
I have tried to highlight text and use ctrl-c to copy and it does not work, also ctrl-v is not working in normal mode and i have to run it twice to work, don't know if this is the expected behavior or not.

but ctrl-c is not working neither in normal nor insert mode and to copy something i have to highlight it then right click and copy with mouse.

Regards

morenoh149:
Refs #541

### DIFF
--- a/lib/settings.coffee
+++ b/lib/settings.coffee
@@ -12,7 +12,7 @@ settings =
       default: false
     useClipboardAsDefaultRegister:
       type: 'boolean'
-      default: false
+      default: true
     numberRegex:
       type: 'string'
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
> fixes #541

I don't think this fixes that issue. That issue is about the `ctrl-c` and `ctrl-v` key-bindings (which map to `copy` and `paste` on Linux and Windows) being overridden by this package. Still, I think this is a good change. Thanks @morenoh149.

Issue #334, opened 2014-07-03
remapping Esc


assemnical:
I remapped  'jk' and 'kj' to activate command mode.
 
>'.editor.vim-mode:not(.command-mode)':
  'j k': 'vim-mode:activate-command-mode'
  'k j': 'vim-mode:activate-command-mode'

However if I just press 'k' or 'j' followed by another letter, the 'k' or 'j' doesn't appear. How can I overcome this problem?

BallisticPain:
I have been unable to get jj to work for typing "j" and also using "jj" to get me out of command mode. :(

I feel this shouldn't be so difficult. I have seen people reference a "timeout", but have been unable to find this anywhere in the settings for either Vim Mode nor Atom.

Thoughts?

Thanks!
J

pelachile:
@BallisticPain where did you find the misspelling? I have tried and tried and I can't get jj to work. It's the only thing that prevents me from using atom, I'm just to used to my vim settings.

BallisticPain:
@pelachile, I'll copy and paste below what I have that's working. I use jj constantly and it was imperative that I got it working as well so I fully understand your frustration.

I personally mis-spelled a portion of the ~/.atom/init.coffee script. I didn't copy and paste, I just typed it in.
```
# Your init script
#
# Atom will evaluate this file each time a new window is opened. It is run
# after packages are loaded/activated and after the previous editor state
# has been restored.
#
# An example hack to make opened Markdown files always be soft wrapped:
#
# path = require 'path'
#
# atom.workspaceView.eachEditorView (editorView) ->
#   editor = editorView.getEditor()
#   if path.extname(editor.getPath()) is '.md'
#     editor.setSoftWrap(true)
#

atom.workspaceView.command 'insert-incomplete-keybinding', (e)->
    if oe = e.originalEvent && e.originalEvent.originalEvent
        char = String.fromCharCode(oe.which)
        char = char.toLowerCase() unless oe.shift
        atom.workspace.activePaneItem.insertText(char)
```

I misspelled the .workspace portion.

Here's my ~/.atom/keymap.cson (I was originally modifying the one in the package directly, and this wasn't working as well as modifying my home directory version)...
```
# Your keymap
#
# Atom keymaps work similarly to stylesheets. Just as stylesheets use selectors
# to apply styles to elements, Atom keymaps use selectors to associate
# keystrokes with events in specific contexts.
#
# You can create a new keybinding in this file by typing "key" and then hitting
# tab.
#
[...]

PR #946, opened 2016-01-07
Document a tip for improved Insert Mode mappings for ESC

rdlugosz:
As discussed at length in issue #334, mapping double characters to ESC
is a common Vim pattern (e.g, 'jj' or 'jk'). Atom supports this style of
mapping, but the editor behavior is inconsistent with Vim in that the
initial characters that partially match a mapping are not displayed on
screen. This is jarring to the user.

This PR adds documentation to the README that describes a two-step
approach to mapping ESC that removes the delay from the editor and much
more closely imitates Vim's behavior.

This issue exists today in Atom 1.3.2. A future update to Atom may
change the behavior of
[KeymapManager](https://atom.io/docs/api/v1.3.2/KeymapManager) so that
partial matches are displayed. Work in this area was most recently done in
[PR #94 on atom-keymap](https://github.com/atom/atom-keymap/pull/94).
It's unlikely that changing this behavior is a high priority, so I suspect this
technique will be valuable for a long time to come.

I considered adding this info to a Wiki, but since this project doesn't
currently have one I figured the README was the next-best place.
Remapping ESC is sufficiently common among Vim users that the info
should be right up front.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,50 @@ which works in conjunction with this plugin.
 
 Currently, vim-mode has some issues with non-US keyboard layouts. If you are using a keyboard layout which *isn't* American and having problems, try installing [keyboard-localization](https://atom.io/packages/keyboard-localization).
 
+#### Double-key Keymaps in Insert Mode
+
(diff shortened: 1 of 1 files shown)


lee-dohm:

PR #967, opened 2016-02-17
Add Z Z command to close pane

natesholland:
This is an initial stab at getting the 'Z Z' command to work.
Normally in vim this would save and close the window. In this implementation
it still asks you for a save dialog box.

I would ideally like it to not prompt with the dialog box but that is definitely beyond my skill level right now. I think this is better than nothing, but I also understand if other's want to have a closer match to vim before we merge this in. I would love some feedback and tips on how to get this working without the pop up box asking you to save.

### DIFF
--- a/keymaps/vim-mode.cson
+++ b/keymaps/vim-mode.cson
@@ -182,6 +182,7 @@
   'ctrl-w q': 'core:close'
   'g t': 'pane:show-next-item'
   'g T': 'pane:show-previous-item'
+  'Z Z': 'pane:close'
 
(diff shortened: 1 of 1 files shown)


dcalhoun:
@natesholland if you desire to implement this without the prompt, I'd recommend reviewing the [vim-mode-zz](https://github.com/sQu1rr/vim-mode-zz/blob/master/lib/vim-mode-zz.coffee) package. It currently does what you desire. You could attempt to port it's abilities into vim-mode package.

PR #968, opened 2016-02-21
Fix some scroll bugs

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
Fix #961

PR #970, opened 2016-02-24
Fix a scroll when contents are wrapped

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
In the document that has many wrapped contents, scroll handling commands (ctrl-d, ctrl-f and ctrl-e) can NOT move screen to end of document.

This PR refers to #968.
To explain the problem more details, I uploaded a video file.
[video reproduce the problem](https://youtu.be/cu8EDDZt_aU)

I hope it is helpful to check and review my PR.
Thank you.

### DIFF
--- a/lib/motions/general-motions.coffee
+++ b/lib/motions/general-motions.coffee
@@ -471,14 +471,20 @@ class ScrollKeepingCursor extends Motion
 
   execute: (count) ->
     # TODO: remove this conditional once after Atom v1.1.0 is released.
-    if @editor.setFirstVisibleScreenRow?
-      newTopRow = @getNewFirstVisibleScreenRow(count)
-      super(count)
-      @editor.setFirstVisibleScreenRow(newTopRow)
-    else
-      scrollTop = @getNewScrollTop(count)
-      super(count)
-      @editorElement.setScrollTop(scrollTop)
+    # TODO: Atom v1.1.0 was released.
+    # But setFirstVisibleScreenRow() use getLineCount() not
+    # getScreenLineCount(). Once after it use getScreenLineCount(),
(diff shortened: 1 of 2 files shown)


lib/motions/general-motions.coffee
```diff
@@ -471,14 +471,20 @@ class ScrollKeepingCursor extends Motion
...
-      super(count)
-      @editor.setFirstVisibleScreenRow(newTopRow)
-    else
-      scrollTop = @getNewScrollTop(count)
-      super(count)
-      @editorElement.setScrollTop(scrollTop)
+    # TODO: Atom v1.1.0 was released.
+    # But setFirstVisibleScreenRow() use getLineCount() not
+    # getScreenLineCount(). Once after it use getScreenLineCount(),
+    # [1. by setFirstVisibleScreenRow] can be changed to [2. by scrollTop]
```
maxbrunsfeld: > But setFirstVisibleScreenRow() use getLineCount() not getScreenLineCount()

Thanks so much for catching this! I think that we should just fix that method. There are some issues with using the `TextEditorElement::getScrollTop`, because it doesn't always update until the next animation frame.

maxbrunsfeld: @b6pzeusbc54tvhw5jgpyw8pwz2x6gs Are you interested in submitting a PR to atom/atom fixing this? I can publish a hotfix release of Atom so that we don't need to wait for 1.7.0 to come out.

b6pzeusbc54tvhw5jgpyw8pwz2x6gs: @maxbrunsfeld , ok I will PR to atom/atom.

maxbrunsfeld:
Ok, I believed the atom/atom PR referenced above. Thanks @b6pzeusbc54tvhw5jgpyw8pwz2x6gs!

PR #991, opened 2016-03-29
Copy active item on split

viddo:
This was the behavior until a few days ago (sometime after updating to Atom 1.6.0).

I have not been able to track down what change caused the change, but noticed that the [default split behavior](https://github.com/atom/atom/blob/420a8d8692080a1d067caa936f6888ff6653a2f7/keymaps/darwin.cson#L86-L89) is slightly different, so is there any reason for why not change the keymapping here to match the default keymaps behavior?


lee-dohm:

Issue #1010, opened 2016-04-19
Repeat last change does not repeat insertion


mveres:
- go to insert mode
- type 'hello'
- go to normal mode
- navigate do a different row
- press "."
Expected: 'hello' is inserted again
Actual: it isn't

movestill:
Probably related, trying to repeat a change word command (cw) only deletes the current word but does not do the subsequent insert of text.

Atom 1.7.1
vim-mode 0.65.0
Windows 7

psifertex:
Looks like a dupe of #1006

haasDev:
+1

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
I have the same issue.

Atom 1.7.1
vim-mode 0.65.0
ubuntu 14.04

sandman21dan:
+1 using command CIW then repeating with "." doesn't re-add the word, just deletes.

snkutlu:
Have the same issue for "Insert" and for "change-word". 

Atom 1.7.3
vim-mode 0.65.0 
MacOS ElCapitan 10.11.4

gjngeldenhuis:
+1

gregz67:
+1

Same atom and vim-mode versions. Mac Yosemite.

BinaryMuse:
Fixed by #1020

Issue #1006, opened 2016-04-14
repeat no longer repeats inserts after a deletion


sunjay:
One of the awesome things I used to be able to do to quickly replace a word is type `cwFoo` and then use `.` to repeat that everywhere else. Today I restarted atom and installed some updates. This is no longer working now.

### Here's a complete example:

Let's say I have the following lines:

```
var import = {};
import.a = 1;
import.b = 3;
```

If I go to the `i` of the first line, I expect that typing `cwexport<Esc>` will replace `import` with `export`. Going to the beginning of the second and third line and repeating that command with `.`, I expect that each of the other `import`s will be replaced with `export`.

This would look like this:

```
var export = {};
export.a = 1;
export.b = 3;
```

Currently, running those exact commands results in this:

```
var export = {};
.a = 1;
.b = 3;
```

The deletion gets repeated, but the insert doesn't. That would be fine if I was using `d`. With `c`, I expect the insertion to be part of the repeat. 

This works as expected in regular vim.

sunjay:
Any update on this? This is still happening.

scotte:
Same issue here - "." is a pretty fundamental and important vi binding...

horiuchi:
I have same issue too.

Atom 1.7.2
vim-mode 0.65.0
MacOSX

jessejanderson:
Same issue here, as @scotte said this is a pretty fundamental binding...

Atom 1.7.1
vim-mode 0.65.0
Mac OS X El Capitan

jessejanderson:
Co-worker exhibits the same issue with Atom 1.7.0 and vim-mode 0.63.0

morenoh149:
seeing this as well. No recent commits here seem related https://github.com/atom/vim-mode/commits/master could it be an interaction between installed packages?

sunjay:
Doesn't seem like it. Only vim-mode is acting on that key.

![ss 2016-04-21 at 03 36 58](https://cloud.githubusercontent.com/assets/530939/14721904/fbfe1804-07d6-11e6-908d-7c84290bc7c5.png)

morenoh149:
Don't get how the latest commit was on march 10th and this wasn't reported until April 14th though...

sunjay:
Maybe something in the core broke it?

morenoh149:
then maybe 1.7.0 or 1.7.1 broke it

https://github.com/atom/atom/releases/tag/v1.7.0

[...]

PR #1020, opened 2016-05-10
:bug: Fix repeat-insertion bug referenced in #1006

stefanvlaski:
Hi all,

I took a shot at fixing the repeat-insertions-after-change bug referenced in #1006 and #1010. As @t9md pointed out, there is an issue with `item.confirmChanges(changes)` function of the `Insert` class. That function constructs an instance of `TransactionBundler` which relies on `change.newRange` and some other properties which don't exist. As a result, `bundler.buildInsertText()` always returns `""`, which is why the dot-command after `cwText` repeats the deletion, but not the insertion.

This fix makes use of the `getChangesSinceCheckpoint` function, implemented in atom/text-buffer@9f9025fd7dbac7b1bf2ebc58bf1176df9d9f251c. That function replaces the function of the same name implemented in `lib/vim-state.coffee`, which uses private API's. It essentially supplies equivalent information in a different format. The `TransactionBundler` class is changed accordingly.

The dot-command now behaves as expected on conventional patterns like `iText`, `cwText`, `c2wText`, `ccText`, etc.

There is some funky behavior when moving the cursor outside of the edited region while in insert-mode and continuing to type (not when fixing something that was just typed). However, I don't think this is a result of this fix, but rather a consequence of the way `TransactionBundler` bundles the inserted text by calling `@editor.getTextInBufferRange`. Similar behavior occurs using the release version of the vim-mode package on the 1.6.2 Atom release, when the dot-command was still working. To illustrate, take

`One two three.`

In command-mode, move the cursor to the beginning of `two`. Type `cwHello` which replaces `two` by `Hello`. Without leaving insert-mode, move the cursor to the beginning of the line and type `world.<space><escape>`. I see

`world. One Hello three.`

Now move the cursor to the beginning of the word `three` and hit `.`. I see in

Atom 1.7.3 + this fix:
`world. One Hello world. One Hello.`
All of `world. One Hello` is taken to be the typed text, and hence re-inserted after hitting `.`. Everything between insertions is grouped together and taken to be typed text, even if it was there before.

Atom 1.6.2 + vim-mode 0.65.0
`world. One Hello d. One H.`
This time `d. One H` is taken to be the typed text and reinserted after hitting `.`.

(Neo)vim
`world. One Hello world. three.`
In vim, it seems like `cw` is forgotten about once the cursor leaves the original region. `.` simply repeats the last insert (`world.<space>`) and does not delete `three`.

Since this kind of pattern isn't really the intended use of the `cw`, it's hard to say what the behavior should be, but the difference to vim is worth noting.

Feedback is greatly appreciated.

### DIFF
--- a/lib/operators/input.coffee
+++ b/lib/operators/input.coffee
@@ -13,8 +13,10 @@ class Insert extends Operator
   isComplete: -> @standalone or super
 
   confirmChanges: (changes) ->
-    bundler = new TransactionBundler(changes, @editor)
-    @typedText = bundler.buildInsertText()
+    if changes.length > 0
+      @typedText = changes[0].newText
+    else
+      @typedText = ""
 
   execute: ->
     if @typingCompleted
@@ -134,33 +136,6 @@ class Change extends Insert
       @typingCompleted = true
     else
       @vimState.activateNormalMode()
-
-# Takes a transaction and turns it into a string of what was typed.
-# This class is an implementation detail of Insert
-class TransactionBundler
-  constructor: (@changes, @editor) ->
-    @start = null
-    @end = null
-
-  buildInsertText: ->
-    @addChange(change) for change in @changes
-    if @start?
-      @editor.getTextInBufferRange [@start, @end]
-    else
-      ""
-
-  addChange: (change) ->
-    return unless change.newExtent?
-    if @start?
-      if @start.isGreaterThanOrEqual(change.start)
-        @start = change.start
-    else
-      @start = change.start
-    if @end?
-      if @end.isLessThanOrEqual(change.start.translate(change.newExtent))
-        @end = change.start.translate(change.newExtent)
-    else
-      @end = change.start.translate(change.newExtent)
 
 
 module.exports = {
--- a/lib/operators/input.coffee
+++ b/lib/operators/input.coffee
@@ -150,55 +150,17 @@ class TransactionBundler
       ""
 
   addChange: (change) ->
-    return unless change.newRange?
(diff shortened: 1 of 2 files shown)


stefanvlaski:
### Update:
After Travis threw some failed tests when repeating insertions with multiple cursors, I did some more digging. `buffer.getChangesSinceCheckpoint` already bundles changes in a reasonable manner, making the `TransactionBundlerClass` redundant. In fact, the way I had implemented it, `TransactionBundler` overbundled changes, which caused the issues with inserting with multiple cursors.

Commit @e13e028 removes the `TransactionBundler` class completely and grabs the bundled changes directly from `buffer.getChangesSinceCheckpoint`. This resolves the failed Travis checks in relevant tests. The build is still not successful, but that is caused by three failures in unrelated tests, which are present in the master branch as well.

This change also has a side-effect on the previously mentioned funky behavior. I now get in Atom 1.7.3 + this fix:
`world. One Hello world. .`
The `buffer.getChangesSinceCheckpoint` bundles the two insertions `Hello` and `world.<space>` separately, allowing us to fetch only the most recent bundle and executing `cwworld.<space>` on the `.` command. This is a step closer to the vim behavior (`iworld.<space>`). Fully emulating vim behavior here (if desired) should probably be a separate PR, since it is related to the operator history, not the text itself.

BinaryMuse:
@stefanvlaski Thanks very much for taking this on!

jacekkopecky:
Thanks @stefanvlaski !

PR #1048, opened 2016-08-04
Make paragraph forward match Vim's behavior re: whitespace.

ewannema:
:bug: Consecutive empty lines should not be treated as different paragraphs and
whitespace only lines are not paragraph boundaries.

http://vimdoc.sourceforge.net/htmldoc/motion.html#paragraph

I went in to fix the failing test on forward paragraph behavior and noticed that vim-mode also didn't match Vim for whitespace only line behavior. If this PR makes sense I can also take a look at backwards paragraph motion.


ewannema:
I temporarily updated the regular expression to handle \r\n line endings. I modified the test on my machine to validate it, but I am not sure how the project wants to handle these tests.

Issue #1045, opened 2016-07-25
No longer updated?


v3ss0n:
Is this package is dead? vim-mode-plus is successor ?

DrakeXiang:
+1

plbowers:
I, too, would like to know if this package is dead...

I've been working hard on a few modifications/enhancements, but if there's low probability of them getting merged into the master trunk (it looks like the last merge happened in March?) then I'm not sure I want to keep working on them...

v3ss0n:
total of 39 PRs.. would be a shame if this package is dead  , someone should take over.

maxbrunsfeld:
We're not developing new features on this package right now, though we still want to fix regressions and keep it working. A lot of people now use vim-mode-plus which is very well maintained.

v3ss0n:
ok then checking out vim-mode-plus

plbowers:
If this is the official status of this package then it REALLY needs to be made clear. From a quick read it looked like vim-mode-plus was a breakaway fork and this was still the standard-bearer for vim within atom. Now I've spent a few weeks learning the package and implementing some new features and now I find out that vim-mode-plus is not a "rogue breakaway" but the official new standard-bearer. Please make this clear as it will make a BIG difference to people deciding between the 2 packages...

plbowers:
It should also be notated at https://atom.io/packages/vim-mode unless the README automatically gets displayed there...

lee-dohm:
As stated in [the README](https://github.com/atom/vim-mode), this package is no longer maintained and is deprecated. We recommend that people use the [vim-mode-plus package](https://atom.io/packages/vim-mode-plus) instead. Because of this, we are [archiving this repository](https://help.github.com/articles/about-archiving-repositories/) and closing all issues and pull requests. Thanks very much for your support and contributions!

PR #1070, opened 2016-10-01
Document that package is deprecated

plbowers:
On this comment:

https://github.com/atom/vim-mode/issues/1045#issuecomment-250221562

It was recommended that vim-mode-plus be used in place of vim-mode. This PR documents that so new users will know to go there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,11 @@
 
 Provides Vim modal control for Atom, blending the best of Vim and Atom.
 
-### Current Status
+### Current Status - DEPRECATED in favor of vim-mode-plus
+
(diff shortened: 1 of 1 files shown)


50Wliu:
@maxbrunsfeld

bronson:
I really hope this can be merged.  A number of issues people have filed against vim-mode in the past two months are already fixed in vim-mode-plus.

Everyone should know to try vim-mode-plus before they spend much time tracking and filing vim-mode issues.

maxbrunsfeld:
Thanks @plbowers.

PR #1083, opened 2016-12-02
Add deprecation notice to package description

50Wliu:
This builds on #1070 and changes the package description to make it clear that vim-mode-plus should be installed instead.  After this is merged a new patch should be released.

We should also consider moving this repository to the atom-archive organization.

/cc @maxbrunsfeld @t9md


maxbrunsfeld:
Thanks @50Wliu!

